PR #566, opened 2024-11-10
Update leaderboard

arktoswb:
- change columns
- naming: leaderboard -> entries


calebjacob (approved):

calebjacob:
For context: removing the hardcoded `live_bench/average` column will now default to the `important_columns` returned by the API. This will match our `/evaluations` page.

PR #579, opened 2024-11-15
Various agent runner UI fixes

calebjacob:
- Fixed a regression related to our `latest` changes where an agent would open our permissions modal to ask for permission to run itself again if using `latest` instead of fixed version. User won't be prompted now.
- Fixed more iframe height issues that popped up recently thanks to more CSS edge cases.


hub/demo/src/lib/entries.tsx
```diff
@@ -119,6 +119,10 @@ export function idForEntry(entry: z.infer<typeof entryModel>) {
   return `${entry.namespace}/${entry.name}/${entry.version}`;
 }
 
+export function idMatchesEntry(id: string, entry: z.infer<typeof entryModel>) {
+  return id.startsWith(`${entry.namespace}/${entry.name}`);
```
gabehamilton, line 123: Maybe we need the / in here to make it more precise. Though it would be an interesting feature to have a "" named assistant agent to allow permissions for any agent in a namespace.

gabehamilton (approved):

hub/demo/src/lib/entries.tsx
calebjacob: Good point! I'll push a commit here to add `/` at the end.

PR #586, opened 2024-11-26
add confidential openai chat client

0xshawn:
Add an example of OpenAI chat client with [PHALA Confidential AI API](https://docs.phala.network/confidential-ai-inference/confidential-ai-api) in TEE. The function of listing models will be added in latter PR.

TODO
- [ ] List models
- [ ] Verify the attestation


arktoswb:
Is this still relevant? I assume it has never been reviewed because there are no reviewers.

PR #591, opened 2024-12-02
Version Fixes

calebjacob:
- Add .nvmrc file to UI for consistent local dev node/npm version
- Enforce Node and NPM version usage. This mostly helps prevent someone using an older version of `npm` that would completely rewrite our `package-lock.json` file based on versioning/format changes.
- Remove trailing slash on unused NEXT_PUBLIC_BASE_URL env var to avoid future confusion


calebjacob:
I removed the Python version changes after realizing why 3.12 can't be supported right now. The `.nvmrc` and env var change for `NEXT_PUBLIC_BASE_URL` are still worth merging.

gabehamilton (approved):

PR #594, opened 2024-12-02
Get vector store file details by filename

gabehamilton:


PierreLeGuen:
When is this needed?

gabehamilton:
When working with complex agents like https://github.com/nearai/official-agents/tree/main/common-tool-library some regular database operations are needed on files. This and https://github.com/nearai/nearai/pull/414 allow some of those operations without the agent author having to run a separate database.

PierreLeGuen (approved):

Issue #592, opened 2024-12-02
"Add user secret" dialog


zavodil:
It would be great to add to `agent-iframe-requests.ts` new feature:

`nearai_add_secret` -> that is generating `nearai_add_secret_response`

potential input:
```
export const agentAddSecretRequestModel = z.object({
  agentId: z.string(), // full line like namespace/agent/version
  key: z.string(),
  value: z.string(),
  requestId: z.string().nullish(),
  reloadAgentOnSuccess: z.bool().default(true), // working title
});
```

and for `nearai_add_secret_response` to return a status of the operation. 

After adding a secret it is expected for agent to have it in its env variables, so would be nice to immediately reload this agent within the current thread, if `reloadAgentOnSuccess` is set to True.

if agentId != currentAgentId, we have to warn a user that this may affect other agent.

It may be useful to notify user if a given secret already exists and ask to update it, but probably we can avoid this extra complexity, user has an option to remove a secret manually and try to add secret again.

PR #595, opened 2024-12-02
Support "add_secret" iframe post message flow

calebjacob:
Closes: https://github.com/nearai/nearai/issues/592

- If the user allows the secrets to be added, any conflicting secret key will be removed before the new value is added (any previous value will be overwritten). We can probably polish the UX further overt time here.
- The `Always Allow` option isn't applicable for `add_secrets` since we always want to notify the user when an agent is adding a secret. The UI simply shows an `Allow` button if the agent is only adding requesting to add secrets.
-  Went with `add_secrets` and `add_secrets_response` action names to handle the case where an agent wants to add multiple at once.
- The iframe is reloaded if `reloadAgentOnSuccess` is true (which defaults to `true`) after the secrets were added

Request via `add_secret` action:

```ts
const agentAddSecretsRequestModel = z.object({
  secrets: z
    .object({ agentId: z.string(), key: z.string(), value: z.string() })
    .array(),
  requestId: z.string().nullish(),
  reloadAgentOnSuccess: z.boolean().default(true),
});
```

Response via `add_secret_response` action:

```
{
  action: 'add_secrets_response';
  requestId?: string;
  result: {
    addedKeys: string[];
    failedKeys: string[];
  };
}
```

### DIFF
--- a/hub/demo/package-lock.json
+++ b/hub/demo/package-lock.json
@@ -10,7 +10,7 @@
       "dependencies": {
         "@gera2ld/tarjs": "^0.3.1",
         "@hookform/resolvers": "^3.6.0",
-        "@near-pagoda/ui": "^2.1.1",
+        "@near-pagoda/ui": "^2.1.2",
         "@near-wallet-selector/bitte-wallet": "8.9.14",
         "@near-wallet-selector/core": "8.9.14",
         "@near-wallet-selector/here-wallet": "8.9.14",
@@ -827,9 +827,9 @@
       }
     },
     "node_modules/@near-pagoda/ui": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/@near-pagoda/ui/-/ui-2.1.1.tgz",
-      "integrity": "sha512-wnTTG5voPuj19xm+nTzq8UKj4p9lObqGG1Ui+3+cuX2xr3u+QJwj6CxgcEIjCkUtGL5lM3NpjbJp8N6ox4+Y1A==",
+      "version": "2.1.2",
+      "resolved": "https://registry.npmjs.org/@near-pagoda/ui/-/ui-2.1.2.tgz",
+      "integrity": "sha512-se0YWwhyvhDD0pILhJNIxUhdP7NA74Ox5lU/ET6U5Mbk/1G0ktaLNXz1kBW9pv0jrMjLQnXwSvdk5dHZhXVc/g==",
       "dependencies": {
         "@phosphor-icons/react": "^2.1.5",
         "@radix-ui/colors": "^3.0.0",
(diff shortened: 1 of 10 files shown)


zavodil (changes requested):
Great job! The core functionality works as a charm.

I notice a few suggestion, that may be can adjusted on this stage.

1) on your screen you have a request to add secret to zavodil.near/ui-agent/latest. Can we always display the exact agent version even when working with the latest version? adding secrets to `latest` seems wrong.

2) What is the proper way to add a secret for all versions of agent? On a backend we have a condition (`version` IS NULL OR version = ''), but adding secret to `a.near/bot/` raised an error for me. Can we enable `a.near/bot/` or something like `a.near/bot/*`?

3) In this PR we have a way for agent to overwrite other agent's secret, but it may happens without user to understand what is going on. This message is too small and not clear.
Screenshot 2024-12-03 at 11 41 46 AM

If malicious agent will insert/overwrite it's own record for other agent, then user be forced to sign malicious action later on. Can we show a visible warning for this case, when targetAgentId != currentAgentId? It may be the valid case to add secrets for other agents but user must understand what is he doing.

4) Scenario 3) allows agent to overwrite secret of other agent and this is dangerous. If secret has a private key, and another agent overwrite it, user is loosing access all funds related to the account associated with the initial key. So for now we can disable an option for agent to overwrite secrets of other agents. We can display an alert like "this agents wants to update a secret of other agent, but you already have a secret there, which may be lost." User has a workaround to do it manually: open that agent and overwrite a value manually or delete a pervious value and try to add a secret again. May be it's a bad UX but I afraid if users will loose private keys and seed phrases this way. Any suggestions of how to make this UX better are highly welcomed.

calebjacob:
Thanks for the review @zavodil - lots of good points. I pushed some UI changes that should improve the UX and make it more clear when external agents are being updated and/or current values are being overwritten.

I haven't implemented any logic yet that actually prevents an external agent secret being overwritten, but the user should clearly be able to see what's going on and make a decision if they want to proceed or not. Let me know if you'd prefer I implement logic that completely prohibits this scenario - wouldn't be hard at all.

I also implemented changes that will resolve this issue:

> What is the proper way to add a secret for all versions of agent? On a backend we have a condition (version IS NULL OR version = ''), but adding secret to a.near/bot/ raised an error for me. Can we enable a.near/bot/ or something like a.near/bot/*?

The iframe can post any of the following formats for `agentId` to save for all versions of an agent:

```
account.near/agent-name
account.near/agent-name/latest
account.near/agent-name/*
```

All of these values will be coerced to `version: ""` when the UI calls the API to create the secret.

Screenshot 2024-12-03 at 1 18 41 PM

Screenshot 2024-12-03 at 1 18 35 PM

Feel free to test further with my changes and let me know if you see any issues.

zavodil:
Thank you, it works great!

PR #597, opened 2024-12-03
Added py-near to lambda runners.

gabehamilton:
also printing param keys and not values in runner logs.


zavodil (approved):

gabehamilton:
> @gabehamilton I added 2 more deps, I hope it's not conflicting with frameworks you created

Looks great. I'll run a test against staging to make sure.

PR #609, opened 2024-12-04
Remove save/download environment endpoints.

gabehamilton:


gabehamilton:
> LGTM, much better to be consistent about using threads vs environment. Can we run threads locally today?

Yeah, a local run can write to a Thread maintained by a local hub or a remote hub.

PR #617, opened 2024-12-05
Handle non flaky benchmark inference errors

arktoswb:
Flaky errors are handled by `num_inference_retries` being 10 by default. However, some time-out errors are non flaky, as been shown by model `qwen2p5-coder-32b-instruct`: ~0.2% of coding queries and higher percentage on non-coding queries (can be as high as 20% on some tasks)

### DIFF
--- a/nearai/solvers/__init__.py
+++ b/nearai/solvers/__init__.py
@@ -100,7 +100,7 @@ class SolverInferenceSession:
                 return response_content
         except Exception as e:
             print(f"Error: {e}")
-            return " "
+            return f"{e}"
 
 
 class SolverStrategy(ABC, metaclass=SolverStrategyMeta):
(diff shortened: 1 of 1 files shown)


nearai/solvers/__init__.py
```diff
@@ -79,24 +79,28 @@ def add_system_message(self, message: str) -> None:
...
+                        messages=self.messages,
+                        temperature=0.0,
+                    ),
+                )
+                response_content = str(cast(List[Choices], completion_response.choices)[0].message.content)
+                self.messages.append({"role": "assistant", "content": response_content})
+                return response_content
+        except Exception as e:
+            print(f"Error: {e}")
+            return " "
```
cmrfrd, line 103: We should return some standard error string.

arktoswb: Solvers don't expect errors to be in model/agent output. " " is the most stable choice in case of inference errors for solvers.

arktoswb: But it will not be logged this way.

arktoswb: Changed to `f"{e}"`

cmrfrd (approved):

PR #618, opened 2024-12-05
Add 'page' into evaluation_table api

arktoswb:


calebjacob:
For our 0.5B competition, what value would you expect to be passed for the `page` param? Will it always be just "competition" or "competition_0.5b"?

arktoswb:
> For our 0.5B competition, what value would you expect to be passed for the `page` param? Will it always be just "competition" or "competition_0.5b"?

"competition_0.5b"

arktoswb:
Deprecated.

PR #622, opened 2024-12-06
Agent docs fix

gabehamilton:


thisisjoshford (approved):
Thanks for the update @gabehamilton !

PR #634, opened 2024-12-10
nearai agent dev command starts local UI

gabehamilton:


zavodil:
I like this idea!

Shall we try to open http://localhost:3000 automatically?

And the choice of name for this command is not entirely clear. Previously, for `nearai agent`, we specified the agent's name (or path), but here we don't do that. Additionally, the command `dev` doesn't seem to associate with launching any kind of UI for me. Perhaps something like `nearai hub` or `nearai ui` would be more intuitive to me.

nearai/cli.py
```diff
@@ -474,6 +475,15 @@ def read_solutions(self, entry: str, status: Optional[bool] = None, verbose: boo
 
 
 class AgentCli:
+    def dev(self) -> None:
+        if not os.path.exists("hub/demo/.env"):
+            shutil.copy("hub/demo/.env.example", "hub/demo/.env")
+
+        ret_val = os.system("npm run dev --prefix hub/demo")
```
calebjacob, line 482: Can this simply be `npm run dev`? I'm not sure what `--prefix hub/demo` is adding here unless I'm totally missing something.

We also need to run `npm i` to install dependencies. Maybe we change to `ret_val = os.system("npm i && npm run dev")`.

Also, the UI requires a minimum of Node 18 - might be worth mentioning somewhere.

I'm not sure how the Python registry works for `pip install nearai`. We'll want to make sure all of `hub/demo`'s source code is published and `nearai dev` works as expected when installed that way.

calebjacob (changes requested):
Nice work! Left one comment to work through.

calebjacob (approved):
Looks good to start with! We can keep making this more robust with future PRs.

PR #635, opened 2024-12-10
fix integration

cmrfrd:


zavodil:
Could you add `py_near` deps too if all conflicts are resolved? Unless it's already added in another PR.

gabehamilton (approved):

Issue #636, opened 2024-12-10
Default Agent is too basic


gagdiez:
The default agent only says "Hello, world" and does not reply to any interaction with the terminal

We can easily make it more interactive by simply uncommenting these lines: 

https://github.com/nearai/nearai/blob/3fc906f0c6d4e115a64f0bc5ad986dbe102462fe/nearai/cli.py#L722-L728

This will drastically reduce the "time to first wow"

gagdiez:
closed by #631

PR #637, opened 2024-12-10
Make default agent interactive

gagdiez:
The current agent only replies "Hello, World" to any prompt given by the user. This makes it look like `nearai` agents are broken, since the user will see a lot of things happening in the console, but they will always get the same answer

In this commit, I have modified the default agent so it is interactive by default

This effectively allows us to greatly reduce the "time to first wow" by a lot in the quickstart tutorial, besides not making the `nearai` look like it is broken

This tackles issue #636

### DIFF
--- a/nearai/cli.py
+++ b/nearai/cli.py
@@ -720,12 +720,10 @@ class AgentCli:
 
 
 def run(env: Environment):
-    env.add_message("assistant", "Hello, world!")
     # Your agent code here
-    # Example:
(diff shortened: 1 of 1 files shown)


gabehamilton:
Great minds think alike. In my PR yesterday I updated the quickstart to use `nearai agent create` and thought it was strange to have these lines commented out. I updated them as well since `env.add_reply` is the newer method for agents to use.
I think you can close this PR.

gagdiez:
Happy to close this given that there is a merged solution

PR #639, opened 2024-12-10
Reorganize Docs - First Major Pass

gagdiez:
Our current Quickstart tutorial was too concise when explaining how to quickstart an agent, and too lengthy when explaining how to do other things.

So far, I have divided the current `agent.md` into 3 files:

- `agents/quickstart.md` which only explains how to get a local agent and how to use it
- `agents/registry.md` which explains how the agent registry works
- `agent.md` which contains the info that I have still need to migrate

---

- [x] Create the quickstart
- [x] Create the agent registry
- [x] Explain the agent environment
- [x] Explain environmental variables


gagdiez:
This PR assumes that #642 and #641 are both fixed, otherwise, commands in the docs will not work (though, to be fair, they are not working now)

docs/agents/quickstart.md
```diff
@@ -0,0 +1,171 @@
...
+NearAI agents can communicate with each other, and have access to a wide range of capabilities, including
+authentication, tools, apis, smart contract calls, and more.
+
+Let's build our first agent on Near AI!
+
+---
+
+## Pre-Requisites
+
+To get started, you will need to have a [Near account](https://wallet.near.org/), and install the [Near AI CLI](https://github.com/nearai/nearai/#setup):
```
charleslavon, line 15: You may want to point them directly to a wallet that works well with NearAi, like Bitte. As some of the other ones do not support the required 'SignMessage' functionality

charleslavon: ah, nevermind :)  you mention the same suggestion below

gagdiez:
Preview: https://docs-preview.onrender.com/

docs/agents/env/messages_files.md
```diff
@@ -0,0 +1,113 @@
...
+### Storing Data
+
+To create a new file in the thread we can use the `write_file` method from the environment:
+
+```python title="agent.py"
+def run(env: Environment):
+  env.write_file('file.txt', 'hello thread')
+```
+
+??? tip "Where is the file stored?"
```
gabehamilton, line 63: This is static now as of 0.1.6. `/tmp/nearai-agent-runner/gagdiez.near/hello-ai/1.0.0` for instance should have files.

docs/agents/env/tools.md
```diff
@@ -0,0 +1,57 @@
...
+    tool_registry.register_tool(my_tool)
+    tool_registry.register_tool(my_tool2)
+    response = env.completions_and_run_tools(messages, tools=tool_registry.get_all_tool_definitions())
+```
+
+---
+
+## Terminal Commands
+
+Agents have access to the local terminal through the environment, the following methods are available:
```
gabehamilton, line 52: Future enhancement. We could call out that it is manually written agent code where this methods can be called. To provide them to the LLM, the author needs to use the tool registry and completions_and_run_tools.

docs/index.md
```diff
@@ -1,204 +1,65 @@
-# Near AI
+# NEAR AI: Building a Truly Open AI
 
-See the [Near AI website](https://near.ai) for more on how we will achieve _Open Source and User-owned AGI._.
+Welcome! [NEAR AI](https://near.ai) is a toolkit to help build, measure, and deploy AI systems focused on [agents](./agents/quickstart.md).
 
+Driven by one of the minds behinds **TensorFlow** and the **Transformer Architecture**, with NEAR AI, your data is yours,
+and your AI works for your. No compromises on privacy or ownership.
```
gabehamilton, line 6: 'works for you', 'works for your benefit' ?

docs/login.md
```diff
@@ -1,64 +0,0 @@
-# NEAR Login
```
gabehamilton, line 1: These more advanced login cases still seem relevant. Can this be moved somewhere?

gabehamilton (approved):
Only concern is the login.md page.

docs/agents/env/messages_files.md
gagdiez: I have `0.1.6` installed

I can see `chat.txt` and `system_log` in `/tmp/nearai-agent-runner/`

But the files created by my agent (`env.write_file`) are being stored in `/private/var/folders/v6/pw6s7x0n6z9d05hwvnfq0w5h0000gn/T/agent_7c81c486f08d45708e0131dc215ef98d`

docs/agents/env/tools.md
gagdiez: I have not arrived to this file yet, so it is verbatim how it was originally written. I'll get to it this week most probably, so maybe we could leave it for a future improvement

docs/login.md
gagdiez: I put them on the `quickstart`, and was planing to bring them on every (future) quickstart we create
 
https://github.com/nearai/nearai/blob/b9f39067c23e14f78d9ea4b77f315674763cd610/docs/agents/quickstart.md?plain=1#L76-L86

I don't think having a separate `login.md` file is necessary (as many people will only use the normal login flow), but if you prefer to have it back, I have no problem on adding it somewhere

let me know

gabehamilton (approved):

Issue #274, opened 2024-09-09
Forking an agent in frontend


ilblackdragon:
Make an easy way to fork an agent and modify code/metadata.

I would imagine "Fork" button in https://app.near.ai/agents/near.ai/agent/1.0.0 screen and some screen with editing files like agent.py and metadata.json.

PR #653, opened 2024-12-12
Fork registry entry

calebjacob:
Closes: https://github.com/nearai/nearai/issues/274

- Fork any registry entry (not just limited to agents)
- Add `Develop` button on entry details page to open instruction modal

### DIFF
--- a/hub/api/v1/registry.py
+++ b/hub/api/v1/registry.py
@@ -18,7 +18,7 @@ from sqlmodel import col, delete, select, text
 
 from hub.api.v1.auth import AuthToken, get_auth, get_optional_auth
 from hub.api.v1.entry_location import EntryLocation, valid_identifier
-from hub.api.v1.models import RegistryEntry, Tags, get_session
+from hub.api.v1.models import Fork, RegistryEntry, Tags, get_session
 
 DEFAULT_NAMESPACE_WRITE_ACCESS_LIST = [
     "spensa2.near",
@@ -309,6 +309,11 @@ def list_files_inner(entry: RegistryEntry) -> List[Filename]:
     return files
 
 
+class ForkOf(BaseModel):
+    namespace: str
+    name: str
+
+
 class EntryInformation(BaseModel):
     id: int
     namespace: str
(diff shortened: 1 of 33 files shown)


hub/demo/src/server/api/routers/hub.ts
```diff
@@ -415,6 +416,59 @@ export const hubRouter = createTRPCRouter({
...
+            name: input.name,
+            version: input.version,
+          },
+        }),
+      });
+
+      if (response.status === 409) {
+        throw new TRPCError({
+          code: 'CONFLICT',
+          message: 'Test asdf',
```
gabehamilton, line 452: Update?
Since you're finding the conflicting resource server side you could pass through the name and skip the client side lookup. Fine to just put a placeholder in here though since it's all written.

gabehamilton (approved):

hub/demo/src/components/lib/Code.tsx
```diff
@@ -16,6 +15,7 @@ import {
 import s from './Code.module.scss';
 
 export type CodeLanguage =
+  | 'bash'
```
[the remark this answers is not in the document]
calebjacob: Good catch - `shell` is more generic. Updated: https://github.com/nearai/nearai/pull/653/commits/48de90fa0304c1b9d430b65d27753384f6a0cdbe

hub/demo/src/server/api/routers/hub.ts
calebjacob: That's a good point. This additional throw logic isn't really needed due to the client side validation catching 99% of the issues with better UX. I simplified things here: https://github.com/nearai/nearai/pull/653/commits/50c505e55b6fd4d77b49d7d6a6126ef5fd9146b3

hub/api/v1/registry.py
```diff
@@ -498,3 +498,64 @@ def list_entries_inner(
...
+
+        return {
+            "status": "Entry forked and uploaded",
+            "entry": {
+                "namespace": new_entry.namespace,
+                "name": new_entry.name,
+                "version": new_entry.version,
+                "category": new_entry.category,
+            },
+        }
```
mfornet, line 561: I suggest adding a return type for this call so we can have the API type checked and the API generator can pick up the result from this method.

You can see how it is done on other methods.

mfornet (approved):
LGTM in general, I checked briefly over the frontend side of the changes.

hub/api/v1/registry.py
calebjacob: Good idea - updated this here: https://github.com/nearai/nearai/pull/653/commits/3ad45a8723c5b18d5bece017d0832c8b1372be52

PR #655, opened 2024-12-13
feat: hub to read commands in EVENT_JSON logs in NEAR blocks and run …

zavodil:
…corresponding agents


hub/api/v1/neardata.py
```diff
@@ -0,0 +1,224 @@
...
+
+@asynccontextmanager
+async def lifespan(app):
+    # remove last known block_id to reset the state on start
+    if RESET_BLOCK_ID_ON_START:
+        if os.path.exists(BLOCK_ID_FILE):
+            os.remove(BLOCK_ID_FILE)
+            print(f"File {BLOCK_ID_FILE} has been deleted.")
+
+    auth_token = load_auth_token()
```
gabehamilton, line 215: Is this what you are planning to change to a 'Hub auth token'? Even though it would be convenient to load it from config I think we should load if from somewhere else since this is a common area of confusion. It would be easy for a developer to write functionality that uses the hub auth when what they need is the user's auth.

hub/api/v1/neardata.py
```diff
@@ -0,0 +1,224 @@
+import json
```
gabehamilton, line 1: This file could go in `hub/tasks` instead of `hub/api/v1`

hub/api/v1/neardata.py
zavodil: Yes, that's a good point. Yes, this is the Hub auth token. If we separate it from the general config, are you suggesting creating a separate file and instructions on how to set up a dedicated config? I feel that might be more complicated than adding a section to the instructions on how to run the local hub, where we can clarify the purpose of the token in this case.

gabehamilton: Maybe the Hub token can go in `hub/.env` since it is the hub's auth and therefore be distinct from user auth in the config.json? The .env.example can refer to it, the 'how to run a local hub' docs can explain how to create one. We can then add a `sign_as_hub` function to neardata.py or x_event_source that pulls it from the environment and uses it..

gabehamilton (approved):

hub/api/v1/neardata.py
zavodil: Please take a look, I just added docs to clarify it for agents builders and hub maintainers.

PR #660, opened 2024-12-18
Feat Twitter Event source

gabehamilton:
See also test agent https://github.com/nearai/official-agents/pull/4


hub/tasks/x_event_source.py
```diff
@@ -0,0 +1,164 @@
...
+# Path to store the last read tweet id
+TWEET_STATUS_FILE = "last_tweet.txt"
+
+scheduler = AsyncIOScheduler()
+
+# todo pull from registry 'integration' item
+EVENT_SOURCE_CONFIG = {"user_id": "1867270324649160704"}
+
+# todo pull from agent metadata
+LISTENER_CONFIG = {"agent_name": "flatirons.near/nearsecretagent/0.0.1"}
```
zavodil, line 24: I assume the best dev experience is to set a field `trigger_by_users_mentions`:  `user_id` (or maybe even [`user_id1`, `user_id2`]) in the metadata and here to select all of them from the database

hub/tasks/x_event_source.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+# Function to get the last read tweet timestamp from a file
+async def get_last_tweet_timestamp():
```
zavodil, line 28: Shall we do it user specific? I assume we will have agents will the different frequency of mentions read.

hub/tasks/x_event_source.py
```diff
@@ -0,0 +1,164 @@
...
+
+    run_params = RunCreateParamsBase(
+        assistant_id=agent,
+        model=DEFAULT_MODEL,
+        instructions=None,
+        tools=None,
+        metadata=None,
+        include=[],
+        additional_instructions=None,
+        additional_messages=[{"content": message, "role": "user"}],
```
zavodil, line 107: It might be help helpful to provide tweet_id too, so agent will read extra context itself if needed.
so "content": json.dumps({"message": ...,  "tweet_id": ..})

gabehamilton: Yeah, was working on this data format in the agent. Also will remove this function and use the one in nearevents

hub/tasks/x_event_source.py
gabehamilton: Yeah, maybe `triggers: twitter: mentions: user_id`. In any case, event sources and trigger config can go in another PR.

hub/tasks/x_event_source.py
gabehamilton: I think we want to use this to get all tweets since the last time we fetched tweets. I haven't implemented that though since I was testing that the agent handled duplicate tweets.

hub/tasks/x_event_source.py
```diff
@@ -0,0 +1,81 @@
...
+
+RESET_BLOCK_ID_ON_START = True  # reusing setting name for now
+
+# Path to store the last read tweet id
+TWEET_STATUS_FILE = "last_tweet.txt"
+
+scheduler = AsyncIOScheduler()
+
+# future: pull from registry 'integration' item
+EVENT_SOURCE_CONFIG = {"user_id": "1867270324649160704"}
```
zavodil, line 20: Please move it unless to .env, put I prefer to have a final implementation to read user_ids from registry before to merge

zavodil (changes requested):

hub/api/v1/agent_data.py
```diff
@@ -0,0 +1,78 @@
...
+
+class AgentDataRequest(BaseModel):
+    namespace: str
+    name: str
+    key: str
+    value: Union[dict[Any, Any]]
+
+
+def is_hub_account(auth: AuthToken):
+    return True  # todo implement
```
zavodil, line 24: Please update this, since we have a functionality to handle Runner Api Key and `is_trusted_runner_api_key()` method

zavodil (changes requested):
Do you mind to update
def is_hub_account(auth: AuthToken):
    return True  # todo implement
too?
Now we have a functionality to handle Runner Api Key and is_trusted_runner_api_key()

zavodil (approved):

PR #662, opened 2024-12-18
feat: hub to derive a private key for every agent and generate a veri…

zavodil:
feat: hub to derive a private key for every agent and generate a verifiable signature for every hub completion

feat: enable 3 methods in agents environment:
- signed_completion, to return {response, signature, public_key}
- verify_signed_message, docs to be added ASAP
- get_agent_public_key 

fix: `agents` object in a runner's service restricted to a single agent only

How to test:
in `.env` to add:
```
HUB_PRIVATE_KEY="ed25519:PROPER_MASTER_PRIVATE_KEY...." 
TRUSTED_RUNNER_API_KEYS=["abc123"]
```

for runner add a param to docker run:
```
docker run -e RUNNER_API_KEY=abc123
```

agent to try: `zavodil.near/verifiable_agent/0.1`
(be sure to run in against docker runner, not in the local mode)


nearai/aws_runner/Dockerfile
```diff
@@ -1,14 +1,36 @@
...
 
+CMD /bin/sh -c "echo Setting agent runner user as $AGNER_RUNNER_USER"
+
+RUN /usr/sbin/groupadd -r $AGNER_RUNNER_USER && \
+    /usr/sbin/useradd -r -g $AGNER_RUNNER_USER -m $AGNER_RUNNER_USER
+
+RUN mkdir -p /$AGNER_RUNNER_USER/.nearai/registry && \
+    chown $AGNER_RUNNER_USER:$AGNER_RUNNER_USER /$AGNER_RUNNER_USER/.nearai/registry
+
+COPY nearai/aws_runner/.env $HUB_CONFIG_PATH
```
gabehamilton, line 22: Since the `nearai` folder gets copied in on line 33 we'll need to move aws_runner back out of nearai (which we should do anyway).

zavodil: Thank you, do you believe https://github.com/nearai/nearai/pull/662/commits/6f64ce7a5117601a045e9bead07ca77180e98066 is resolving it?

hub/api/v1/sign.py
```diff
@@ -0,0 +1,65 @@
...
+    payload = CompletionSignaturePayload(
+        agent_name=agent_name,
+        completion=response_message_text,
+        model=model,
+        messages=messages_json,
+        temperature=float(temperature),
+        max_tokens=int(max_tokens),
+    )
+
+    agent_key = derive_new_extended_private_key(hub_private_key, agent_name)
```
gabehamilton, line 57: Interesting that `latest` and the underlying version if specified directly will have different keys. Seems like a feature for now though.

gabehamilton (approved):

PR #669, opened 2024-12-30
Set release to use python 3.11

gabehamilton:
The release to PyPi https://github.com/nearai/nearai/actions/workflows/release.yml has been failing https://github.com/nearai/nearai/actions/runs/12439186816/job/34732522578

Since it appears to be using 3.12 and it's only recently that our dependency tree prevented 3.12 from building, it seems like a likely culprit.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,7 +62,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: '3.x'
+        python-version: '3.11'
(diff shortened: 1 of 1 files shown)


PierreLeGuen:
What's the motivation behind this change?

gabehamilton:
> What's the motivation behind this change?

Thanks, I updated the description.

cmrfrd (approved):
3.12 is a little too new to be supporting, capping at 3.11 makes sense

PR #670, opened 2025-01-02
fix: protected auth

zavodil:


nearai/shared/inference_client.py
```diff
@@ -30,244 +30,274 @@
...
-        if config.auth is not None:
-            self._auth = config.auth.generate_bearer_token()
-        else:
-            self._auth = None
-        self.client = openai.OpenAI(base_url=self._config.base_url, api_key=self._auth)
-
-    # This makes sense in the CLI where we don't mind doing this request and caching it.
-    # In the aws_runner this is an extra request every time we run.
-    # TODO(#233): add a choice of a provider model in aws_runner, and then this step can be skipped.
-    @property
```
PierreLeGuen, line 42: I believe removing the `@property` will remove caching, is it expected?

gabehamilton (approved):

PR #678, opened 2025-01-07
Simplify details.agent.x

calebjacob:
- Simplify `details.agent.x` to `details.x` while being backwards compatible
- Remove unused details keys in UI


gabehamilton:
LGTM. We just need matching backend changes (with backwards compatibility) before we can merge this.

gabehamilton:
I don't know when we'll get to the server side. Maybe close for now and we can pull it back out of storage if the subject bubbles back up?

PR #690, opened 2025-01-09
Cleanup use of local runner on server

gabehamilton:


PierreLeGuen:
What if I have hub server running locally and I want to run agents locally? Does this change prevent this?

gabehamilton:
If the use case is local UI talking to local Hub, no custom runner, config of `RUNNER_ENVIRONMENT=local_runner`; yes this would prevent that case. 

Is that a case we want to support? I haven't been testing that case. If it is, we can just close this PR. 

If it's not a configuration we want to support then this PR would remove that feature, thus reducing the number of configurations we need to test with new runner features and changes.

PR #693, opened 2025-01-09
Updated poetry.lock

gabehamilton:
* Fixes poetry.lock for main
* Updated boto versions to reduce dependency search time.


nearai/config.py
```diff
@@ -4,7 +4,7 @@
 from typing import Any, Callable, Dict, Optional
 
 import openai
-import urllib3
+import urllib3  # type: ignore
```
ewiner, line 7: This isn't a good way to solve the issue.

Looking at the `poetry.lock` file, `urllib3` is coming in as a dependency of `boto3`, using a new version for Python 3.10 and above and 1.x for Python 3.9.  For 1.x, type definitions are available on typeshed (`types-urllib`).  So I think you should add a 3.9-only dependency:
```
types-urllib3 = { version = "^1.26.25.14", python = "<3.10" }
```
(made ☝️ by hand, haven't verified it)

Putting a `type: ignore` on there is problematic because first of all we lose the typing help for that part of config, but more importantly if there's breaking changes between the `urllib3` versions now we've lost the typecheck on that too and would be more likely to break 3.9 compat. 

(or we drop 3.9 compat entirely, which is a bit on the early side for a library but not crazy as it hits EOL in October)

ewiner (dismissed):

nearai/config.py
zavodil: Unfortunately poetry 2 doesn't support carets in project.dependencies https://python-poetry.org/docs/dependency-specification/#caret-requirements

ewiner: Oh the caret isn't the important part, I just wrote that out of habit.  The version constraint is what matters.

gabehamilton: Ah, nevermind, it's because of the boto version upgrades: must have pulled in a different urllib3 than the one previously pulled in by the openapi client.

cmrfrd (approved):
Passes lock checks, and without the typing ignore, I think this pr is fine

PR #697, opened 2025-01-11
Local agent cache and inference client (cached provider models) cache.

gabehamilton:
* Store compiled code in Agent
 * Local cache for agents
 * Local cache for inference client - to cache fetched providers

### DIFF
--- a/nearai/aws_runner/service.py
+++ b/nearai/aws_runner/service.py
@@ -274,7 +274,12 @@ def start_with_environment(
     )
     global inference_client
     global inference_client_cache_time
+    # Force a check for new models after an hour if the runner has stayed hot for that long
+    # this is a failsafe given usage patterns at the time of writing, if models are uploaded more frequently and
+    # runners stay hot, it could be adjusted down or client model caching removed.
     if not inference_client or (time.time() - inference_client_cache_time > 3600):
+        if inference_client_cache_time:
+            write_metric("InferenceClientCacheCleared", "1", "Count")
         inference_client = InferenceClient(client_config, protected_vars.get("RUNNER_API_KEY"), agent.identifier)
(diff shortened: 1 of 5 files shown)


nearai/aws_runner/service.py
```diff
@@ -255,7 +272,13 @@ def start_with_environment(
         base_url=api_url + "/v1",
         auth=auth,
     )
-    inference_client = InferenceClient(client_config, protected_vars.get("RUNNER_API_KEY"), agent.identifier)
+    global inference_client
+    global inference_client_cache_time
+    if not inference_client or (time.time() - inference_client_cache_time > 3600):
```
zavodil, line 277: Could you comment in the code why this 1 hour interval was chosen?

nearai/aws_runner/service.py
```diff
@@ -135,6 +145,11 @@ def load_agent(client, agent, params: dict, additional_path: str = "", verbose=T
         agent_metadata = client.get_agent_metadata(agent)
         stop_time = time.perf_counter()
         write_metric("GetMetadataFromRegistry_Duration", stop_time - start_time, verbose=verbose)
+        full_agent = Agent(
```
zavodil, line 148: There is a problem here if you cache a '/latest' agent version, and then dev uploads new code for his agent, and updates the `latest`, then cache will be invalid. 

so we need a workaround for "/latest" here. But I assume it's a good time to remove "/latest" from the runner at all and only keep it on a UI/agent caller.

zavodil (changes requested):

nearai/aws_runner/service.py
gabehamilton: Good catch. Yes, I'll change the runner invocation in the api to pass the actual version.

PR #703, opened 2025-01-13
feat: run_scheduler

zavodil:
feat: run_scheduler. persistent scheduler where user can create an agent run in hours / days / weeks


hub/api/v1/schedule_run.py
```diff
@@ -0,0 +1,62 @@
...
+from hub.api.v1.models import RunSchedule, get_session
+from hub.api.v1.models import Thread as ThreadModel
+
+schedule_run_router = APIRouter(tags=["Run Schedule"])
+
+logger = logging.getLogger(__name__)
+
+
+class CreateScheduleRunRequest(BaseModel):
+    """Request model for creating a new hub secret."""
```
gabehamilton, line 19: Needs update

gabehamilton:
Looks like this duplicates functionality in thread_routes with the schedule_at param. https://github.com/nearai/nearai/blob/main/hub/api/v1/thread_routes.py#L490C5-L490C16 

Do you plan to remove that param in thread_routes?

zavodil:
> Do you plan to remove that param in thread_routes?

From my understanding at this stage we still can have two. May be for some persistent hub runner in TEE, just a thread_routes param will be enough. New run_scheduler can be switched off in this case.

hub/tasks/schedule.py
```diff
@@ -25,9 +100,16 @@ def load_auth_token():
 async def lifespan(app):
     """Schedule recurring tasks."""
     auth_token = load_auth_token()
+    run_scheduler = os.getenv("RUN_SCHEDULER", "false").lower() == "true"
```
gabehamilton, line 103: I think this could be READ_SCHEDULED_RUNS to match the others but more importantly so that it doesn't sound like it turns the whole scheduler on and off.
```suggestion
    run_scheduler = os.getenv("READ_SCHEDULED_RUNS", "false").lower() == "true"
```

hub/alembic/versions/26e1e353eb58_create_run_scheduler_table.py
```diff
@@ -0,0 +1,37 @@
...
+# revision identifiers, used by Alembic.
+revision: str = "26e1e353eb58"
+down_revision: Union[str, None] = "c5af2948a246"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "schedule_runs",
```
gabehamilton, line 23: I think `scheduled_runs` would be clearer. Once the come into this table they have been scheduled.
```suggestion
        "scheduled_runs",
```

gabehamilton (approved):

PR #709, opened 2025-01-15
feat: format verbose printed statements

thisisjoshford:
- Formats verbose printed statements to end-users when running agents and allows content to be more digestible. 
- Adds `[DEBUG]` to statements to indicate `verbose` mode that can be changed in a future update :) 

### Before
Screenshot 2025-01-14 at 9 46 42 PM

### After
Screenshot 2025-01-15 at 12 48 31 PM

### DIFF
--- a/nearai/aws_runner/service.py
+++ b/nearai/aws_runner/service.py
@@ -151,7 +151,8 @@ def load_agent(client, agent, params: dict, additional_path: str = "", verbose=T
                         f"    Tags: {', '.join(agent_metadata['tags']) if agent_metadata['tags'] else 'None'}\n"
                         f"    Model: {agent_metadata['details']['agent']['defaults']['model']}\n"
                         f"    Model Provider: {agent_metadata['details']['agent']['defaults']['model_provider']}\n"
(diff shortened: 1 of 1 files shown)


arktoswb (approved):

gabehamilton (approved):

zavodil:
This implementation is wrong. I see errors and agent doesn't run

```
  File "/Users/alice/projects/jasnah-cli/nearai/aws_runner/service.py", line 166, in load_agent
    [DEBUG]   • Model: {agent_metadata['details']['agent']['defaults']['model']}
KeyError: 'defaults'
```
@arktoswb @gabehamilton @thisisjoshford

thisisjoshford:
> And please ensure there’s a way to opt out of verbose mode from the CLI before making it so extensive. I was only able to manage it by hardcoding `params` in `cli.py`

I wasn't adding more here, just formatting what was already printed out to the end-user. I was actually asking @gabehamilton how to disable this as I didn't find a straightforward way to change `verbose` to `false` and we agreed that the CLI should have this flag. 

Thanks for catching the error and quick fix btw!

Issue #705, opened 2025-01-14
Share auth session between app.near.ai and chat.near.ai


calebjacob:
Currently, if you sign in to `app.near.ai` and then visit `chat.near.ai` you will need to sign in again. This is a confusing, frustrating user experience for most people.

Instead of saving the token from `auth.near.ai` in local storage, we should save it in our Next JS middleware as a `Secure`, `HttpOnly` cookie configured with the `Domain=near.ai` attribute (which will allow the cookie to be read on the root and all subdomains of `near.ai`).

PR #714, opened 2025-01-15
Shared auth session between app.near.ai and chat.near.ai

calebjacob:
Closes: https://github.com/nearai/nearai/issues/705

- Ended up needing to clean up our tRPC + React Query configuration to properly support reading and setting cookies in our tRPC middleware layer.
- Migration logic to keep users signed in. This moves their auth token from local storage to a `HttpOnly`, `Secure` cookie scoped to `Domain=near.ai` in production. This will make the cookie accessible for `near.ai` and `*.near.ai` domains.
- ENV variable updated for `app.near.ai` and `chat.near.ai` => `AUTH_COOKIE_DOMAIN=near.ai`
- All other environments (staging, preview builds) won't have a value set for `AUTH_COOKIE_DOMAIN` - meaning the cookie will be scoped only to the exact domain its running on.
- Added a little README for future reference: `hub/demo/src/trpc/README.md`

I've done my best to thoroughly test all the edge cases I can think of and it seems to work smoothly.


hub/demo/src/components/Footer.tsx
```diff
@@ -46,7 +46,7 @@ export const Footer = ({ conditional }: Props) => {
             size="text-xs"
             color="sand-11"
           >
-            near.ai
+            NEAR.AI
```
arktoswb, line 49: Should it be "NEAR AI"?

arktoswb (approved):

hub/demo/src/components/Footer.tsx
calebjacob: Good catch - fixed!

PR #718, opened 2025-01-16
replace python command with python3

Amira1502:
**Description:**

- Updated all instances of `python` to `python3` in code snippets, commands, and instructions within the documentation.

### DIFF
--- a/docs/agents/quickstart.md
+++ b/docs/agents/quickstart.md
@@ -20,7 +20,7 @@ To get started, you will need to have a [Near account](https://wallet.near.org/)
 === "pip"
 
     ``` bash
-    python -m pip install nearai
+    python3 -m pip install nearai
     ```
 
 === "local"


gagdiez (approved):

zavodil:
I don’t understand the point of making such PRs. I would prefer not to receive notifications about them.

gagdiez:
this user reported on our telegram an error while following the docs

using `python3` instead of `python` solved the error for them, and I assume it will solve it for everyone carrying older python installs

zavodil:
@gagdiez This is configurable within the system, right? If another user without the python3 command but with python writes, we're unlikely to switch it back, are we? Besides, our documentation and README consistently use python instead of python3. I just checked the TensorFlow and PyTorch repositories, and their READMEs also simply mention `python`.

gagdiez:
it was my understanding that `python` now always comes with `python3` for compatibility, but if it is not commonly used we can then rollback this no problem

We can simply add a message saying `hey, if this fails, check your python version` and point to the docs explaining how to change the version

Issue #719, opened 2025-01-16
Vector Store Returns the LEAST similar elements


gagdiez:
We currently implement `vector_stores` through our own API, which uses SQL to check for distances within some embeddings

I was checking doing requests on a medium size `vector_store` created from our documentation, and kept getting wrong answers all the time.

Turns out we are ordering the results in `DESC` order (i.e. largest value first, lowest value last) and them `LIMIT`ing the result:

https://github.com/nearai/nearai/blob/main/hub/api/v1/sql.py#L483-L484

This means that we get the X **more distant** results, i.e. the results that are the LEAST similar to the query

PR #720, opened 2025-01-16
Order Vector Store Results in Ascending Order

gagdiez:
When querying the `vector_store` embeddings, we are ordering the results in DESC order (i.e. largest value first, lowest value last), this means that we get the X **more distant results**, which are the results that are the **LEAST similar** to the query

I have removed the `DESC` keyword from the ordering since `ORDER BY` sorts the records in ascending order by default [1]

This PR fixes https://github.com/nearai/nearai/issues/719

[1] https://www.w3schools.com/sql/sql_orderby.asp

### DIFF
--- a/hub/api/v1/sql.py
+++ b/hub/api/v1/sql.py
@@ -480,7 +480,7 @@ class SqlClient:
         SELECT vse.file_id, vse.chunk_text, vse.embedding <-> %s AS distance
         FROM vector_store_embeddings vse
         WHERE vse.vector_store_id = %s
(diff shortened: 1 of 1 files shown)


PierreLeGuen:
ASC is correct, we are doing an euclidean distance (https://docs.singlestore.com/db/v8.9/reference/sql-reference/vector-functions/euclidean-distance/).
If DOT product <*> DESC is better.

PierreLeGuen (approved):

PR #726, opened 2025-01-17
Various hub fixes

zavodil:
fix: x feed duplicate runs 
fix: neardata timeout error handling
fix: store secrets for a specific version 
fix: add agent env_vars in thread run
feat: show the number of attached secrets
fix: agent namespace setter
fix: get_agent_entry cleaned up from missing argument
fix: redundant debug output cleanup
fix: set agent's code as property

### DIFF
--- a/hub/api/v1/thread_routes.py
+++ b/hub/api/v1/thread_routes.py
@@ -603,6 +603,7 @@ def _run_agent(
             f"for run: {run_id} on thread: {thread_id}. Signed by {auth.account_id}."
         )
 
+        # TODO#733 Optimization with using agent.identifier
         specific_agent_version_entry = get_agent_entry(specific_agent_version_to_run, data_source)
         # read secret for every requested agent
         if specific_agent_version_entry:
--- a/hub/api/v1/thread_routes.py
+++ b/hub/api/v1/thread_routes.py
@@ -580,8 +580,6 @@ def run_agent(
 def _run_agent(
     thread_id: str, run_id: str, background_tasks: Optional[BackgroundTasks] = None, auth: AuthToken = Depends(get_auth)
 ) -> OpenAIRun:
-    logger.info(f"Running agent for run: {run_id} on thread: {thread_id}")
-
     with get_session() as session:
         run_model = session.get(RunModel, run_id)
         if run_model is None:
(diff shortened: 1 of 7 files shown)


hub/api/v1/thread_routes.py
```diff
@@ -662,7 +671,8 @@ def _run_agent(
             print(
                 f"Running function {function_name} with: "
                 f"assistant_id={run_model.assistant_id}, "
-                f"thread_id={thread_id}, run_id={run_id}"
+                f"thread_id={thread_id}, run_id={run_id}, "
+                f"user_secrets:{len(user_secrets)}, agent_secrets:{len(agent_secrets)}"
```
gabehamilton, line 675: I don't think we should print these, that'll put them in the cloudwatch logs.

hub/api/v1/thread_routes.py
```diff
@@ -592,19 +590,35 @@ def _run_agent(
...
+            if agent_entry
+            else run_model.assistant_id
+        )
+
+        logger.info(
+            f"Running agent {specific_agent_version_to_run} "
+            f"for run: {run_id} on thread: {thread_id}. Signed by {auth.account_id}."
+        )
 
+        specific_agent_version_entry = get_agent_entry(specific_agent_version_to_run, data_source)
```
gabehamilton, line 606: Optimization: I think the only place latest is set is in the agent.identifier. So if you want the entry location to match we could just change the identifier and save the extra call here.

hub/tasks/near_events.py
```diff
@@ -34,12 +34,20 @@ async def async_fetch_json(url: str) -> Any:
...
-        else:
-            return None
+    try:
+        async with httpx.AsyncClient() as client:
+            response = await client.get(url, headers=headers, timeout=3)
+            response.raise_for_status()
+            if response.status_code == 200:
+                return response.json()
+    except httpx.HTTPStatusError as exc:
+        print(f"HTTP error occurred: {exc}")
```
gabehamilton, line 44: Since it runs every second maybe we should make it more explicit? I'm thinking the logs could get interleaved with other request errors.
```suggestion
        print(f"HTTP error occurred in near_events source: {exc}")
```

hub/tasks/near_events.py
```diff
@@ -34,12 +34,20 @@ async def async_fetch_json(url: str) -> Any:
...
+    except httpx.RequestError as exc:
+        print(f"Request error occurred: {exc}")
```
gabehamilton, line 46:
```suggestion
        print(f"Request error occurred in near_events source: {exc}")
```

hub/tasks/near_events.py
```diff
@@ -34,12 +34,20 @@ async def async_fetch_json(url: str) -> Any:
...
+    except httpx.TimeoutException as exc:
+        print(f"Timeout error occurred: {exc}")
```
gabehamilton, line 48:
```suggestion
        print(f"Timeout error occurred in near_events source: {exc}")
```

gabehamilton (approved):

hub/api/v1/thread_routes.py
zavodil: I prints only length of the object, not the content. Otherwise it was impossible to debug missing data.

hub/api/v1/thread_routes.py
zavodil: Could you make a pr for this optimisation please? I didn't find `agent.identifier` in that method to use to properly follow your idea. I've created a https://github.com/nearai/nearai/issues/733

Issue #707, opened 2025-01-15
Feature Request: Add support for URL parameter to auto-launch agents with the specified message


zavodil:
I would like to request a feature that enables creating a link that automatically launches an agent with a predefined message if the `runWithInitialMessage` (working title) URL parameter is present. This feature would allow for seamless user interactions by preconfiguring agent actions and providing users with a tailored experience.

Example Use Case:

- A Twitter-based agent wants a user to send payment via the agent at the endpoint: `user.near/pay/1`.
- The agent sends a tweet containing a link:
https://app.near.ai/agents/user.near/pay/1/run?runWithInitialMessage="{\"action\":\"pay-me\",\"amount\":2.5}"  
prompting the user to click the link and pay 2.5 NEAR to unlock a feature
- When the user opens the link, the agent is automatically launched, and a form for 2.5 NEAR payment is generated and displayed to the user. This saves 1 action from the user.

Proposed Implementation:

- URL Parameter Handling: Extend the application to recognize and parse the runWithInitialMessage parameter from the URL.
- Agent Initialization: If the parameter is present, auto-launch the specified agent with the provided message as input.
- Fallback Behavior: If the parameter is absent or malformed, default to standard agent launch behavior.

zavodil:
@calebjacob @gabehamilton what do you think about this feature?
This would give me more freedom in creating agents.

calebjacob:
Sounds like a good idea! Should be really easy to knock out. I'll get to this soon.

gabehamilton:
Definitely a cool feature.
Seems like there is some security risk. Maybe we need a confirmation dialog? We could call the same code as the postMessage run_agent command.

zavodil:
> Maybe we need a confirmation dialog

Yes! @calebjacob would be really good to have a dialog to avoid running agents after url hits from the web crawlers and other potential problems.

Can we show a dialog like "Do you really want to run an agent .. with the input message ...?" And one green button below.

Thank you in advance!

gabehamilton:
Would one time use tokens work for this? 
That way you could set once that you trust an app or agent to generate deep links.
Not something to build in this pass but we could in a future pass.

zavodil:
yes, lets try this in a simple way and as soon as many agents will use this feature, we can evolve

PR #727, opened 2025-01-17
Support ?initialUserMessage query param and prompt user for permission

calebjacob:
Closes: https://github.com/nearai/nearai/issues/707

Example URL: `/agents/flatirons.near/example-travel-agent/latest/run?initialUserMessage=%7B"action"%3A"pay-me"%2C"amount"%3A2.5%7D`

If an `initial_user_message` is defined in the agent's metadata, the UI will now prompt for permission as well. The URL query parameter will always take priority and override any potential value defined in metadata.

Screenshot 2025-01-17 at 10 00 01 AM


zavodil (approved):
Great work!

PR #738, opened 2025-01-22
fix: priority for cached files

zavodil:
fix: local files to support subfolders
feat: support serving binary files from agent's registry folder
feat: env.read_file to support binary files
fix: env.get_agent method removed

Agent to verify: https://app.near.ai/agents/zavodil.near/files/0.1/run (with updated hub only, due to `api_client` update)

### DIFF
--- a/nearai/agents/environment.py
+++ b/nearai/agents/environment.py
@@ -547,10 +547,6 @@ class Environment(object):
 
     # end of protected client methods
 
-    def get_agent(self):
-        """Returns the agent that is being run."""
-        return self._agents[0]
-
     def get_tool_registry(self, new: bool = False) -> ToolRegistry:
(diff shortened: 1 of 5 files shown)


nearai/agents/environment.py
```diff
@@ -715,36 +711,54 @@ def get_agent_temp_path(self) -> Path:
...
+                    break
+
+            if not file_content:
+                # Next check agent file cache
+                # Agent output files & thread files take precedence over cached files
+                file_cache = self.get_primary_agent().file_cache
+                if file_cache:
+                    file_content = file_cache.get(filename, None)
+
+            # Write the file content from the tread or cache to the local filesystem
```
gabehamilton, line 750:
```suggestion
            # Write the file content from the thread or cache to the local filesystem
```

gabehamilton (approved):

PR #749, opened 2025-01-24
Refactor config.py. Makes usage of `nearai` library easier.

arktoswb:
No effect on nearai, but addresses a config update problem when using `nearai` as library.

A problem when using `nearai` as library:

Agent (local run) wants to login to NEAR AI, then proceed with other functions. But CONFIG is static with no straightforward way to update it. The change now introduces a new function `load_config()` and fixes `get_client_config(self)` to use object values. Now, the recommend library usage to get a client config is to call `load_config()` function.

### DIFF
--- a/nearai/config.py
+++ b/nearai/config.py
@@ -117,22 +117,27 @@ class Config(BaseModel):
 
     def get_client_config(self) -> ClientConfig:  # noqa: D102
         return ClientConfig(
-            base_url=CONFIG.nearai_hub.base_url,
-            auth=CONFIG.auth,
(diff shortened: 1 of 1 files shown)


gabehamilton (approved):

gabehamilton:
Maybe the functions other than `load_config` should change to `_` names so the public interface is clearer?

arktoswb:
> Maybe the functions other than `load_config` should change to `_` names so the public interface is clearer?

Yes, will be tracking action items for improving public interface of `nearai` library here: https://github.com/nearai/nearai/issues/752

PR #751, opened 2025-01-24
Initial Agent Interaction and Transaction Protocol

gabehamilton:
This is an initial draft version of the agent to agent protocol. All messages and fields will change.


calebjacob:
@gabehamilton Small change I'd suggest. Let's change the primary `json` key to an object instead of an array of objects:

```json
[
  {
    "type": "json",
    "schema": "https://api.near.ai/protocol.schema.json",
    "json": {
      "request_choice": { ... }
    }
  }
]
```

The `message.content` field is already an array, which could contain a combo of multiple text/json groups:

```json
{
  "content": [
    {
      "type": "json",
      "schema": "https://api.near.ai/protocol.schema.json",
      "json": {
        "request_choice": { ... }
      }
    },
    {
      "type": "json",
      "schema": "https://api.near.ai/protocol.schema.json",
      "json": {
        "request_data": { ... }
      }
    },
    {
      "type": "text",
      "text": {
        "annotations": [],
        "value": "Hell world!"
      }
    }
  ]
}
```

docs/protocols/agent_protocol.md
```diff
@@ -0,0 +1,103 @@
...
+Confirmation requests may be sent to the user in the form of a choice message.
+
+## Request Data Message
+
+## Request Payment Message
+
+## Request Verification Message
+
+## State.json
+To share common state between clients and agents on a thread, the state.json file can be used.
```
ewiner, line 85: How does this interact with the threads-compatible API?  As a client author (or a self-hosted AITP-compatible agent developer), am I just reading messages or is there a whole other filesystem API I have to use now?

docs/protocols/agent_protocol.md
```diff
@@ -0,0 +1,103 @@
...
+    supported: ["request_choice", "request_signature"]
+}]
+```
+
+### List of Client Capabilities in the Agent Protocol
+- `request_choice`: The client has special handling for choice messages. 
+A UI client might display the choice in a rich UI component. An assistant agent might make a choice on behalf of the user.
+- `request_data`: The client can respond to a data request message.
+- `request_payment`: The client can respond to a payment request message with a submit payment message.
+- `request_signature`: The client can sign transactions or messages for verification.
```
ewiner, line 47: That reminds me – `request_signature()` doesn't mean a human is doing the signing, as an intermediate agent could have a key.  My first addition to this spec, once the base is drafted, would be a tool for HITL attestations where the response is signed by the client in addition to the user.

docs/protocols/agent_protocol.md
```diff
@@ -0,0 +1,103 @@
...
+- `state.json`: The client can operate on a shared state.json file. A UI client might display the state in a rich UI component.
+
+### Additional protocol aspects
+These are not communicated in the `client_capabilities` field but are part of the protocol.
+
+ - `action handlers`: Clients can specify that messages are part of the expected flow: making a choice, starting a new product search, etc.
+These are communicated to the agent by prefacing a message with <ExpectedAction/>.
```
ewiner, line 54: More like SuggestedAction, no?

ewiner (changes requested):
I still have plenty more to read... but in general I think this is a great start, but we'll need to get a lot more specific about how these messages integrate into the standards for agent interaction, even if that's happening far from NEAR AI.

calebjacob:
Once we merge and deploy [this PR](https://github.com/nearai/nearai/pull/812), we can reference the JSON schema at `app.near.ai/api/v1/aitp.schema.json`

docs/protocols/agent_protocol.md
```diff
@@ -0,0 +1,103 @@
...
+!!! warning "DRAFT"
+
+    This protocol is under active development and discussion. It will change before publication. Only 
+    develop against it if you are prepared to change your implementation.
+
+
+## Compatibility
+Agent Protocol messages are sent in Messages on Threads. This format is an extension
+of the OpenAI Thread Message format and may or may not be supported by other 
+OpenAI compatible libraries. Specifically it introduces a new message content type: `json`.
```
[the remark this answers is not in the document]
gabehamilton: Ideally not, I've been trying to determine whether we do. Messages can not currently contain JSON directly, they are different than the inference responses that can return `tool_calls`. It's because of tool calls that I thought we may need a json type: when an LLM is constructing a JSON tool call response (especially with LLama) and simultaneously building JSON to go inside one of the arguments, and potentially using JSON data inside that; the escaping gets hairy.

docs/protocols/agent_protocol.md
gabehamilton: Files are attachments on Messages on Threads, both in our current implementation and the OpenAI api. So changes to the state.json file come through as Messages.

docs/protocols/agent_protocol.md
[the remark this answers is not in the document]
ewiner: I'm torn about whether to accommodate the quirks and deficiencies of pre-2025 models in the protocol; that'll go away and it's hard to undo!  But if we see growth in smaller, on-edge models, there's no guarantee they won't have similar problem.

Hmm, I wonder if learning into the concept of [Code Agents](https://huggingface.co/docs/smolagents/conceptual_guides/intro_agents#code-agents) makes sense.  Like a client capability could be "I can run Python code that depends on the `aitp@[schemaversion]` package".

gabehamilton:
New work is in aitp repo

PR #754, opened 2025-01-27
fix: set DEFAULT_TIMEOUT and remove MAX_RETRIES to avoid multiple pai…

zavodil:
…d requests on slow completions

### DIFF
--- a/hub/api/v1/routes.py
+++ b/hub/api/v1/routes.py
@@ -8,6 +8,7 @@ from fastapi import APIRouter, Depends, HTTPException, Query
 from fastapi.responses import JSONResponse, StreamingResponse
 from fastapi.security import HTTPBearer
 from nearai.shared.cache import mem_cache_with_timeout
+from nearai.shared.client_config import DEFAULT_TIMEOUT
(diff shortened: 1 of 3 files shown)


hub/api/v1/routes.py
```diff
@@ -183,7 +184,7 @@ def chat_completions(
         raise HTTPException(status_code=400, detail="Provider not supported") from None
 
     try:
-        resp = llm.chat.completions.create(**request.model_dump(exclude={"provider"}))
+        resp = llm.chat.completions.create(**request.model_dump(exclude={"provider"}), timeout=DEFAULT_TIMEOUT)
```
arktoswb, line 187: We should pinpoint where timeout is not propagated. If we set timeout in `InferenceClient` to be 10 min, why is it still 1 min?
This line seems like a good place for debugging. Maybe print/log what timeout we get from `request.model_dump` and what timeout we pass to `llm.chat.completions.create`

arktoswb (approved):

hub/api/v1/routes.py
arktoswb: Then the question is where `timeout` from `litellm` request is received in hub?

PR #779, opened 2025-01-31
Add type of json to Message

gabehamilton:


gabehamilton:
aitp seems to be working ok so far without this

PR #780, opened 2025-01-31
Subthread communication

gabehamilton:


hub/api/v1/thread_routes.py
```diff
@@ -364,10 +432,19 @@ def list_messages(
     ),
     run_id: str = Query(None, description="Filter messages by the run ID that generated them."),
     auth: AuthToken = Depends(get_auth),
+    include_subthreads: bool = True,
```
gabehamilton, line 435: May change this default in the future, will sync on the front end implementation.

nearai/agents/environment.py
```diff
@@ -25,6 +26,8 @@
 )
 from litellm.utils import CustomStreamWrapper
 from openai import NOT_GIVEN, NotGiven, OpenAI
+from openai.types.beta.auto_file_chunking_strategy_param import AutoFileChunkingStrategyParam
```
gabehamilton, line 29: type_check.sh complained that our chunking classes should be openai. I think they weren't present in the openai version we were initially using but are now.

hub/api/v1/thread_routes.py
```diff
@@ -679,16 +756,24 @@ def _run_agent(
         # with get_session() as session:
         if run_model.parent_run_id:
             parent_run = session.get(RunModel, run_model.parent_run_id)
-
             if parent_run:
+                # check parent_run_id of the parent for loops
+                if parent_run.parent_run_id:
+                    raise HTTPException(
```
PierreLeGuen, line 762: Are we sure of this behavior? Does it mean A awaits B, B awaits C is not possible?

nearai/agents/environment.py
gabehamilton: Nevermind, got it sorted

hub/api/v1/thread_routes.py
gabehamilton: Definitely needs some more sophisticated logic. I put this in as an initial safeguard. I'll write an issue for it. Even the callback here could slip through it with a loop.

PierreLeGuen (approved):

gabehamilton:
Related refactoring PR coming soon.

PR #783, opened 2025-01-31
Return [tool.poetry] to pyproject.toml

arktoswb:
Do not approve.
This appears to be a better fix: https://github.com/nearai/nearai/pull/785


zavodil:
Are you sure you use poetry 2? 
https://python-poetry.org/blog/announcing-poetry-2.0.0/#highlights has a structure that you are updating here

Issue #767, opened 2025-01-28
`messages`.`content` to support utf8mb4


zavodil:
The current implementation of the `messages` table does not fully support utf8mb4 encoding, which is required for handling emojis, special characters, and certain symbols like mathematical formulas. As a result, attempts to insert or retrieve messages containing such characters lead to encoding errors, such as UnicodeDecodeError, or broken data.

The utf8mb4 encoding is necessary to properly store and display emojis and complex symbols, which are common in modern applications.

I brifly tried to fix it: recreated table with full support of utf8mb4 (for database, table and a specific `content` JSON object) and made sure for sqllite to use utf8mb4 - but failed.

So I suspect sqlalchemy is the reason of this issue.

SkidanovAlex:
If the culprit ends up being the database, indeed MySQL (and correspondingly SingleStore, which is protocol-compatible with MySQL) by default assigns collation `utf8_general_ci`,  which doesn't support mb4 in a very weird fashion: it truncates everything that follows the first 4-byte character.

Tables can be converted like this to fix it, if it's indeed the case:

```
ALTER TABLE t1
  DEFAULT CHARACTER SET utf8mb4,
  MODIFY col1 CHAR(10)
    CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL,
  MODIFY col2 CHAR(10)
    CHARACTER SET utf8mb4 COLLATE utf8mb4_bin NOT NULL;
```

(we probably want `utf8mb4_general_ci` as the collation of choice)

(see https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-conversion.html)

zavodil:
@SkidanovAlex Thank you, I tried to update db settings too and IMO my local database supports utf8mb4 (I can insert and read emojies from it using direct queries), but hub/agents are still doesn't support them: (

Do you believe it's enough? `character_set_system` is not that easy to update.

```
singlestore> SHOW VARIABLES LIKE 'char%';
+--------------------------+---------------------------------------------------------------+
| Variable_name            | Value                                                         |
+--------------------------+---------------------------------------------------------------+
| character_set_client     | utf8mb4                                                       |
| character_set_connection | utf8mb4                                                       |
| character_set_database   | utf8mb4                                                       |
| character_set_filesystem | binary                                                        |
| character_set_results    | utf8mb4                                                       |
| character_set_server     | utf8mb4                                                       |
| character_set_system     | utf8                                                          |
[...]

PR #802, opened 2025-02-03
Another unicode fix (3/4-byte)

ewiner:
Another workaound for #767

Someone on Telegram complained that our platform no longer supports Russian.   Russian letters are in the three-byte space of UTF-8, so if our problem is actually SQLAlchemy + 4-byte unicode, we don't have to mess with the 2-byte and 3-byte stuff.  This change allows those characters while still banning those problematic emojis.

Note: I haven't run this code yet in production.  I couldn't even repro the original bug locally.  So there's no way for me to know whether this new regex works, or causes additional issues.

### DIFF
--- a/hub/api/v1/models.py
+++ b/hub/api/v1/models.py
@@ -215,12 +215,12 @@ class Message(SQLModel, table=True):
             ]
         if self.content:
             if isinstance(self.content, str):
-                # Filter out non-utf8 characters
+                # Filter out non-utf8 characters (which shouldn't happen!)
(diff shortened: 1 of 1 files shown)


zavodil (approved):

zavodil:
>> Note: I haven't run this code yet in production. 

Yes this gonna be ok. I noticed same issue with cyrillic characters and put a similar hotfix in my `ts_runner` repo, and updated `web-agent` and `ts` runner. Didn't rebuild the `base` because it takes a lot to rebuild `ts` runner now and I fast in rush to make it done.

PR #806, opened 2025-02-03
set default MAX_RETRIES = 0

zavodil:


gabehamilton (approved):

ewiner (changes requested):
Well... then we're trading occasional double-responses for occasional no-responses.  We know those APIs are flaky, so I don't think that's a good tradeoff.

We'll need some way to disambiguate / consolidate thread responses.  Or... to be really fancy, we store both, and display it as a "pick your preferred response" thing like ChatGPT does when they're A/B testing models.

zavodil:
there is a better fix: https://github.com/nearai/nearai/pull/827

PR #807, opened 2025-02-03
Fix 422 errors on empty messages when calling models with tools in langgraph functions

arktoswb:
Tested successfully on this agent:

```
from langgraph.prebuilt import create_react_agent

from nearai_langchain.orchestrator import NearAILangchainOrchestrator, RunMode

from langchain.agents import Tool

orchestrator = NearAILangchainOrchestrator(globals())
# To continue conversation in local mode:
# orchestrator = NearAILangchainOrchestrator(globals(), thread_id="thread_xxxxxx")


def initialize_agent():
    """Initialize the agent."""
    # Initialize LLM.
    llm = orchestrator.chat_model.chat_open_ai_model

    my_made_up_tool = Tool(
        name="frankenfurtership",
        description="When called, returns something.",
        func=lambda *args, **kwargs: "frankengiven"
    )

    # Create ReAct Agent using the LLM and no tools.
    return create_react_agent(
        llm,
        tools=[my_made_up_tool],
        state_modifier=("You are a helpful agent. Your name is CreateReactAgentExample. You have access to the tool frankenfurtership, which you can call if the user asks you to do so."),
    )


executor = initialize_agent()
env = orchestrator.env

if orchestrator.run_mode == RunMode.LOCAL:
    print("Entering chat mode...")
    user_input = input("\nPrompt: ")
    env.add_user_message(user_input)

messages = env.list_messages()
for chunk in executor.stream({"messages": messages}):
    if "agent" in chunk:
        result = chunk["agent"]["messages"][0].content
    elif "tools" in chunk:
        result = chunk["tools"]["messages"][0].content
    env.add_reply(result)

    if orchestrator.run_mode == RunMode.LOCAL:
        print(result)
        print("-------------------")

env.mark_done()
```

with models `llama-v3p1-70b-instruct` and `qwen2p5-72b-instruct`


nearai/openapi_client/api_client.py
```diff
@@ -313,13 +313,8 @@ def response_deserialize(
                 if content_type is not None:
                     match = re.search(r"charset=([a-zA-Z\-\d]+)[\s;]?", content_type)
                 encoding = match.group(1) if match else "utf-8"
-
-                try:
-                    response_text = response_data.data.decode(encoding)
-                    return_data = self.deserialize(response_text, response_type, content_type)
-                except UnicodeDecodeError:
-                    # binary response
```
arktoswb, line 321: This change is introduced manually in https://github.com/nearai/nearai/pull/737/files#diff-b7dd5257541965913bc4bcac5646279076b7b84b0200f6fe992e8cb944ba6e00

api_client.py is a generated file. So every time, someone wants to generate a client, this change will be reverted.

hub/api/v1/completions.py
```diff
@@ -59,5 +59,22 @@ def get_llm_ai(provider: str) -> OpenAI:
 
 
 class Message(BaseModel):
+    """A chat message."""
+
     role: str
-    content: str
+    content: Optional[str] = " "  # Default to single space
```
ewiner, line 65: Any particular reason you picked `" "` instead of `""`?

arktoswb: Just trying to over-address an issue. I want to see if that helps, then, if it does work, change " " -> ""

PierreLeGuen: Did you try locally?

ewiner (commented):
As a temporary hotfix to reduce our error rate this seems like a fine approach.  But as a fix for the actual problem, I think we need to understand what the client is sending:
* no `content`?
* `content: null`?
* `content: [{text: "first part"}, {text: "second part"}]` (which is valid in the [OpenAI API](https://platform.openai.com/docs/api-reference/chat/create))

And adjust our types and the downstream code accordingly, by either:
1. Handling it before the call to `litellm` (maybe the same way you've done here).
2. Providing a better error message that basically says "sorry we don't _exactly_ support the OpenAI API; `content` is required and must be a string"

Notably, our client SDK allows those more sophisticated types because it borrows the types from `litellm`: https://github.com/nearai/nearai/blob/main/nearai/agents/environment.py#L1216

I'm leaving this as a Comment instead of a Request Changes because I don't want to block merge when I'm out.  I'm 👍 to the idea of merging and deploying this workaround while we figure out what to do with this API.

hub/api/v1/completions.py
arktoswb: Updated and successfully tested locally.

arktoswb:
> As a temporary hotfix to reduce our error rate this seems like a fine approach. But as a fix for the actual problem, I think we need to understand what the client is sending:
> 
> * no `content`?
> * `content: null`?
> * `content: [{text: "first part"}, {text: "second part"}]` (which is valid in the [OpenAI API](https://platform.openai.com/docs/api-reference/chat/create))
> 
> And adjust our types and the downstream code accordingly, by either:
> 
> 1. Handling it before the call to `litellm` (maybe the same way you've done here).
> 2. Providing a better error message that basically says "sorry we don't _exactly_ support the OpenAI API; `content` is required and must be a string"
> 
> Notably, our client SDK allows those more sophisticated types because it borrows the types from `litellm`: https://github.com/nearai/nearai/blob/main/nearai/agents/environment.py#L1216
> 
> I'm leaving this as a Comment instead of a Request Changes because I don't want to block merge when I'm out. I'm 👍 to the idea of merging and deploying this workaround while we figure out what to do with this API.

So far based on my debugging this is what I see:
1. Agent calls langgraph functions `create_react_agent().stream({"messages": messages})`. User asks to call a function.
2. The first response from a model is from `assistant`. It can be "I am going to call `frankenfurtership` function or "Could you please provide an argument for the function `frankenfurtership`" or it can be just empty. Interestingly, empty messages at this step are processed ok (empty message is then added by assistant).
3. Then a tool is called and output from a tool is added with role=`tool`
4. Since the last role is not `assistant` (it's `tool`) the langgraph function is making another completion request.  Last 2 messages: `{'role': 'assistant', 'content': ''}, {'role': 'tool', 'content': 'frankengiven'}`. Error appears at this step or later.
5. If a model wants to call a tool again, the process repeats: first, a message from assistant, then from a role. `{'role': 'user', 'content': 'frankenfurtership'}, {'role': 'assistant', 'content': ''}, {'role': 'tool', 'content': 'frankengiven'}, {'role': 'assistant', 'content': ''}, {'role': 'tool', 'content': 'frankengiven'}` and so on. Called 6 times.

The first empty message from assistant does not produce 422 error. Which probably means that 422 errors appear at the 4th step: passing a completion request with an empty assistant message.

hub/api/v1/completions.py
```diff
@@ -59,5 +59,15 @@ def get_llm_ai(provider: str) -> OpenAI:
...
-    content: str
+    content: str = ""
+
+    @field_validator("content", mode="before")
+    @classmethod
+    def ensure_string_content(cls, v):
+        """Convert None to empty string and ensure content is always a string."""
+        if not v:
+            return ""
+        return str(v)  # Convert any non-None value to string
```
gabehamilton, line 73: For OpenAI compatibility content can also be an array of content_type objects and we will need to use that soon for image_url https://platform.openai.com/docs/api-reference/messages/createMessage

gabehamilton (approved):

PR #820, opened 2025-02-05
Store milliseconds in Message.created_at by using Mysql column declaration (for Singlestore). Fetch messages faster when waiting for a run.

gabehamilton:


calebjacob (approved):
Looks good!

hub/demo/src/components/AgentRunner.tsx
```diff
@@ -177,6 +164,25 @@ export const AgentRunner = ({
...
+
+  const threadQuery = trpc.hub.thread.useQuery(
+    {
+      afterMessageId: thread?.latestMessageId,
+      runId: thread?.run?.id,
+      threadId,
+    },
+    {
+      enabled: isAuthenticated && !!threadId,
+      refetchInterval: isRunning ? 150 : 1500,
```
gabehamilton, line 180: This is the only real change, the rest is just moving the some functions so isRunning is declared before this.

calebjacob:
@gabehamilton Looks like you need to run `npm run pre-commit` to fix some lint issues.

hub/alembic/versions/3cd5c742583c_convert_message_sa_datetime_to_sa_.py
```diff
@@ -0,0 +1,37 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = "3cd5c742583c"
+down_revision: Union[str, None] = "26e1e353eb58"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column("messages", sa.Column("created_at_temp", mysql.TIMESTAMP(fsp=6), server_default=sa.func.now()))
```
ewiner, line 23: But `fsp=6` is the important bit, right?  Did we need to (or do we want to) move from `DateTime` to `Timestamp` too?

Issue #831, opened 2025-02-09
Prefix on embeds introduces extra info


gagdiez:
When creating embeddings for `text_chunks`, we introduce a `prefix`:

https://github.com/nearai/nearai/blob/main/hub/tasks/embedding_generation.py#L221

```py
prefix = "search_query: " if query else "search_document: "
response = await client.embeddings.create(input=prefix + text, model=EMBEDDING_MODEL)
```

I tested using this prefix locally, and it introduces enough "noise" such that it can **change results drastically** when comparing documents with user queries.

In my local test, relevant results when from the 2nd and 3rd position to the 5th and 6th when adding suffixes to the documentation being embedded

gagdiez:
Turns out this is used only on vector stores, because of the specific model we choose for the store

PR #832, opened 2025-02-09
Remove prefix when computing embedding vectors

gagdiez:
This PR removes a prefix that is being prepended to all texts for which we compute an embedding, since in my local tests such prefixes changes drastically the comparisons between "documents' embeddings" and "user query's embedding"

This PR addresses issue #831


PierreLeGuen:
`nomic-ai/nomic-embed-text-v1.5` is trained with those prefixes (`search_document:`, `search_query:`). Would you qualify the answers better without those?

https://docs.fireworks.ai/guides/querying-embeddings-models

gagdiez:
Ah, I see, it is a model specific solution, what if the user wants to use a different model? I will close this PR for now, but will leave the issue around, maybe we could addle the embedding prefix at the vector store level, where we know for sure the model being used, or make the prefix depend on the selected model

PR #833, opened 2025-02-10
feat: debug mode to return python error & python traceback

zavodil:
fix: debug mode to support and "debug" key in various cases

This is how it suppose to work:

Screenshot 2025-02-10 at 4 54 23 PM

### DIFF
--- a/nearai/agents/agent.py
+++ b/nearai/agents/agent.py
@@ -121,7 +121,8 @@ class Agent(object):
         if not self.version or not self.name:
             raise ValueError("Both 'version' and 'name' must be non-empty in metadata.")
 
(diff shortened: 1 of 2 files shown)


zavodil:
@gabehamilton @PierreLeGuen shall we expose traceback? May be error message is enough for debugging the code?

PierreLeGuen:
> @gabehamilton @PierreLeGuen shall we expose traceback? May be error message is enough for debugging the code?

imo, the more the better.

zavodil:
> imo, the more the better.

I'm not sure if displaying AWS lambda paths and internal temp folders we create for agents will actually help developers and not confuse them even more. Look at my screenshot above, traceback looks useless, right?

arktoswb (approved):

PR #834, opened 2025-02-10
fix: Proper unicode handling in JSON content using UnicodeSafeJSON

zavodil:
fix: Introduced `UnicodeSafeJSON` type decorator using LONGTEXT storage
fix: Updated Message model's content field
fix: Singlestore connection string updated to force charset


gabehamilton (approved):
Since we are not querying on json elements within messages, jsonb to longtext seems fine.

Issue #863, opened 2025-02-14
Investigate rendering base64 images in markdown


calebjacob:
From our Telegram channel, someone brought up that they couldn't render base64 images via markdown.

Markdown:

```
### Current State
![Screenshot](data:image/jpeg;base64,/9j/4[....]RRQAUUUUAFFFFABRRRQAUUUUAFFFFABRRRQAUUUUAFFFFAH/2Q==)
```

![Image](https://github.com/user-attachments/assets/e3a4f8df-a4da-4056-be84-2bb5c41b58b9)

Issue #823, opened 2025-02-06
Consumer chat UI improvements


calebjacob:
Copied from Slack via @vanessadespain:

1. Chat history on the left side. Could these labels be generated by the conversation rather than a label corresponding to the Agent? We could set a character count and/or do an ellipses (...) if the label extends beyond the available space.

@PierreLeGuen had originally wrote some logic that would generate a descriptive thread title, but ran into some issues and needed to disable it. I can try to revisit that and see what it'd take to get it working.

2. On the consumer side of Chat, I don't necessarily feel like we need to include the Agents tab and this could be exclusive to the Dev Chat experience. Similarly we probably wouldn't need the "plus" button in top right to add an agent, or the documentation link. All of this to make the consumer facing experience far less overwhelming to a new user. We could still place Docs and/or Agents link in the footer if we want, or just keep our like to [Near.ai](http://near.ai/) in the footer and that landing page will have all of those links for us.

Issue #529, opened 2024-11-05
Fix `topic` generation


PierreLeGuen:
Disabled for now.

Error:

```
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/setup/nearai/.venv/lib/python3.10/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
    await app(scope, receive, sender)
  File "/home/setup/nearai/.venv/lib/python3.10/site-packages/starlette/routing.py", line 75, in app
    await response(scope, receive, send)
  File "/home/setup/nearai/.venv/lib/python3.10/site-packages/starlette/responses.py", line 162, in __call__
    await self.background()
  File "/home/setup/nearai/.venv/lib/python3.10/site-packages/starlette/background.py", line 45, in __call__
    await task()
  File "/home/setup/nearai/.venv/lib/python3.10/site-packages/starlette/background.py", line 30, in __call__
    await run_in_threadpool(self.func, *self.args, **self.kwargs)
  File "/home/setup/nearai/.venv/lib/python3.10/site-packages/starlette/concurrency.py", line 42, in run_in_threadpool
    return await anyio.to_thread.run_sync(func, *args)
  File "/home/setup/nearai/.venv/lib/python3.10/site-packages/anyio/to_thread.py", line 56, in run_sync
    return await get_async_backend().run_sync_in_worker_thread(
  File "/home/setup/nearai/.venv/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 2441, in run_sync_in_worker_thread
    return await future
  File "/home/setup/nearai/.venv/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 943, in run
    result = context.run(func, *args)
  File "/home/setup/nearai/hub/api/v1/thread_routes.py", line 311, in update_thread_topic
    completion = chat_completions(
  File "/home/setup/nearai/hub/api/v1/routes.py", line 170, in chat_completions
    raise HTTPException(status_code=400, detail="Model not supported") from None
fastapi.exceptions.HTTPException: 400: Model not supported

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/setup/nearai/.venv/lib/python3.10/site-packages/uvicorn/protocols/http/httptools_impl.py", line 401, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
[...]

PR #866, opened 2025-02-14
UI chat improvements

calebjacob:
Closes: https://github.com/nearai/nearai/issues/823
Closes: https://github.com/nearai/nearai/issues/529
Closes: https://github.com/nearai/nearai/issues/863

The logic for generating the thread topic can probably be improved, but this seems to work well enough for now from my testing. Longer thread titles will be truncated in the sidebar with an ellipsis if they're too long.

- Allow base64 images via markdown image syntax
- Simplify main nav for `chat.near.ai`
- Generate thread titles via inference when starting a new thread
- Double click on a thread in the sidebar to quickly access the "rename topic" modal

### DIFF
--- a/hub/api/v1/routes.py
+++ b/hub/api/v1/routes.py
@@ -172,16 +172,9 @@ def get_agent_public_key(agent_name: str = Query(...)):
 @v1_router.post("/chat/completions")
 def chat_completions(
     db: DatabaseSession,
-    # request: ChatCompletionsRequest = Depends(convert_request)
-    request: ChatCompletionsRequest,
+    request: ChatCompletionsRequest = Depends(convert_request),
     auth: AuthToken = Depends(get_auth),
 ):
-    request = convert_request(request)
-    """
-        The "request: Depends(convert_request)" logic would not run when chat_completions() is called directly from
-        thread_routes.py generate_thread_topic(). Calling convert_request() directly inside this method instead.
-    """
-
     logger.info(f"Received chat completions request: {request.model_dump()}")
 
     try:
--- a/hub/api/v1/routes.py
+++ b/hub/api/v1/routes.py
@@ -172,9 +172,16 @@ def get_agent_public_key(agent_name: str = Query(...)):
 @v1_router.post("/chat/completions")
 def chat_completions(
     db: DatabaseSession,
-    request: ChatCompletionsRequest = Depends(convert_request),
+    # request: ChatCompletionsRequest = Depends(convert_request)
+    request: ChatCompletionsRequest,
     auth: AuthToken = Depends(get_auth),
 ):
+    request = convert_request(request)
+    """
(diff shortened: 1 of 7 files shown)


hub/api/v1/thread_routes.py
```diff
@@ -394,10 +394,6 @@ def create_message(
         if thread is None:
             raise HTTPException(status_code=404, detail="Thread not found")
 
-        # TODO(#529): Fix topic generation
-        # if not thread.meta_data or not thread.meta_data.get("topic"):
-        #     background_tasks.add_task(update_thread_topic, thread_id, AuthData(**auth.model_dump()))
```
ewiner, line 399: I don't love going from a background update to a synchronous update – can we at least add some general error handling, so if the topic generation fails it doesn't bring down the endpoint?

ewiner (changes requested):

hub/api/v1/thread_routes.py
calebjacob: @ewiner Fair point. A background task probably does make more sense for the backend. I pushed some changes to switch it back to a background task. The user will see a default the thread topic (the name of the agent) for a second or two until the background task finishes and updates the UI with the new topic.

hub/api/v1/routes.py
```diff
@@ -172,9 +172,16 @@ def get_agent_public_key(agent_name: str = Query(...)):
...
     db: DatabaseSession,
-    request: ChatCompletionsRequest = Depends(convert_request),
+    # request: ChatCompletionsRequest = Depends(convert_request)
+    request: ChatCompletionsRequest,
     auth: AuthToken = Depends(get_auth),
 ):
+    request = convert_request(request)
+    """
+        The "request: Depends(convert_request)" logic would not run when chat_completions() is called directly from
+        thread_routes.py generate_thread_topic(). Calling convert_request() directly inside this method instead.
```
ewiner, line 182: That's because calling an API route handler directly from Python code is a really bad idea... this is merely one of the many things that will likely go wrong.

Let's factor out the necessary parts of this function into something that can be called from the topic generator and from this route handler.

calebjacob: @ewiner That sounds like a good plan - makes sense. Feel free to push some commits to this PR to fix it up. Heads down on some other priorities at the moment.

ewiner: OK, pushed https://github.com/nearai/nearai/pull/866/commits/f4f650bc4c001fa4f79b702d45f9f25d4f926665.  Warning: my local dev env is currently busted, so I coded this blind and haven't tested it.  But if it fails in prod, it's not the end of the world either, since it's a background task.

ewiner (approved):
Looks good, but I'd like someone to look at my last commit before we merge.

calebjacob:
@ewiner Thanks for the commit! Everything seems to work well with my local testing. Going to merge this.

PR #885, opened 2025-02-17
Runner caches provider models and not inference client.

gabehamilton:


nearai/shared/inference_client.py
```diff
@@ -58,12 +59,21 @@ def generate_auth_for_current_agent(self, config, agent_identifier):
         else:
             self._auth = None
 
-    # This makes sense in the CLI where we don't mind doing this request and caching it.
+    # This makes sense in the CLI where we don't mind doing this request and caching it.f
```
zavodil, line 62:
```suggestion
    # This makes sense in the CLI where we don't mind doing this request and caching it.
```

zavodil (approved):

PR #889, opened 2025-02-18
chore: adding ic-py dependency to base framework to interact with ICP

jaswinder6991:
- Adding ic-py to the base framework to support interacting with ICP. 
- ic-py is the python client for ICP. 

PS: Although the client is not maintained and last commit was in 2022, it is the only client available at the moment for ICP.

### DIFF
--- a/aws_runner/frameworks/requirements-base.txt
+++ b/aws_runner/frameworks/requirements-base.txt
@@ -16,3 +16,4 @@ ed25519 == 1.5
 tweepy >= 4.14.0
 tenacity == 9.0.0
 mcp
+ic-py


zavodil (changes requested):
Please don't merge this. The base runner is definitely not suitable for this dependency. We will discuss this and I will get back to you later.

jaswinder6991:
> Please don't merge this. The base runner is definitely not suitable for this dependency. We will discuss this and I will get back to you later.

Thanks @zavodil, this came from a [discussion](https://t.me/nearaialpha/1643) here in TG. 

Looking forward to hear from you on what is the best place for such a dependency. In general, I think it would be great if we can support python clients from different chains. So in Near AI agent we can at least do view calls.

PR #854, opened 2025-02-12
feat: local runner to support multiple agents in parallel. Doc: aws_t…

zavodil:
feat: local runner to support multiple agents in parallel. Doc: aws_runner/local/README.md
feat: env.completion_and_get_tools_calls - to make completion and find tool_cals. Agent to test with this PR: https://app.near.ai/agents/zavodil.near/tool_decision_agent/latest
fix: tenacity dependency added to python frameworks

PR #896, opened 2025-02-19
Fix env.completions_and_run_tools

arktoswb:
Agent:

```
from nearai.agents.environment import Environment
def run(env: Environment):
    tool_registry = env.get_tool_registry(new=True)
    prompt = {"role": "system", "content": "You are assistant."}
    response = env.completions_and_run_tools([prompt] + env.list_messages(), tools=tool_registry.get_all_tool_definitions())
run(env)
```

This is a revert of a change in here: https://github.com/nearai/nearai/pull/854/files#diff-d7c24ad303831791e888769f5c7d9613326fe1119e10d85dc1239405a232764b


gabehamilton (approved):
Confirmed bug and fix with agent-tests agent, tool test.

PR #902, opened 2025-02-19
Update hub secrets comment.

gabehamilton:


thisisjoshford (approved):
Thanks for the update on this @gabehamilton! 

With the incoming TEE security solution here, I would then also remove the decryption TODO on L#105.

PR #907, opened 2025-02-19
set mcp version, note masa framework deprecation

gabehamilton:


gabehamilton:
This causes a dependency conflict with py-near over httpx library version: 0.26.0 v 0.27

PR #910, opened 2025-02-19
Add secure OpenAI clients to Environment

arktoswb:
This is needed for langchain libraries. Related: https://github.com/nearai/nearai_langchain/pull/20


ewiner:
It's not very hard to extract the API key from a `SecureOpenAI` instance, so I'm not quite sure what you're hoping to accomplish with this PR.  Wanna hop on a call?

ewiner (approved):
👍 as this uses an existing approach.  We'll talk today about improving that approach across the board.

PR #920, opened 2025-02-20
Fix the thread list when include_subthreads is false

ewiner:
We were briefly showing all threads to everyone!


cmrfrd (approved):

mfornet (approved):

calebjacob:
Whoops! Thanks for catching and fixing.

PR #927, opened 2025-02-21
Remove Max Iterations View

### LINKED ISSUE #891: Remove PARAMETERS from agent UI run.
`max_iterations` parameter currently is not used. It's always set to 1 regardless of UI value. We will deprecate this parameter or will change how it is used.

arktoswb:
Fixes #891


calebjacob (approved):
Thanks for knocking this out @arktoswb - sorry I've had to stay heads down on other tickets. Looks good!

PR #933, opened 2025-02-22
Allow agents to be embedded as an iframe

calebjacob:
Closes: https://github.com/nearai/nearai/issues/931

- Introduce an `/embed/...` route that allows other developers to embed an agent on any website

Wallet sign in via `auth.near.ai` is not possible when embedded as an iframe due to security restrictions set by the wallet providers (which is a good thing).

The developer embedding the iframe will need to integrate with the `auth.near.ai` flow and pass the resulting `#` params as part of the iframe `src`. This will trigger our existing `/sign-in/callback` route to sign in with these URL parameters.

Example embed code:

```tsx
const hash

<iframe
  src="https://app.near.ai/embed/author.near/cool-agent/latest?showThreads=true&showOutputAndEnvVars=true#signature=foobar&accountId=foobar.near&publicKey=ed25519foobar&nonce=00000000000000000001740184404399"
  sandbox="allow-scripts allow-popups allow-same-origin allow-forms"
  style={{
    border: 'none',
    height: '600px',
  }}
/>
```

- `?showThreads=true/false` determines if the left sidebar of threads is shown
- `?showOutputAndEnvVars=true/false` determines if the right sidebar of file outputs and env vars is shown
- The `#` params at the end of the URL will be provided by `auth.near.ai` redirecting back to the developer's website after a successful sign in. This does not include `nonce` - the developer must provide this value to `auth.near.ai` and to the iframe `#` params.

Essentially the iframe `src` URL can be constructed roughly like so:

```js
const nonce = '00000000000000000001740184404399';
const src = `https://app.near.ai/embed/${namespace}/${name}/${version}${window.location.hash}&nonce=${nonce}`;
```

### DIFF
--- a/hub/demo/src/app/embed/[namespace]/[name]/[version]/page.tsx
+++ b/hub/demo/src/app/embed/[namespace]/[name]/[version]/page.tsx
@@ -1,0 +1,41 @@
+'use client';
+
+import { useRouter } from 'next/navigation';
+import { useEffect } from 'react';
+
+import { AgentRunner } from '~/components/AgentRunner';
+import { useEntryParams } from '~/hooks/entries';
+import { useQueryParams } from '~/hooks/url';
+import { SIGN_IN_CALLBACK_PATH, SIGN_IN_RESTORE_URL_KEY } from '~/lib/auth';
+
+export default function EmbedAgentPage() {
+  const router = useRouter();
+  const { namespace, name, version } = useEntryParams();
+  const { queryParams } = useQueryParams([
+    'showThreads',
+    'showOutputAndEnvVars',
+  ]);
+  const showThreads = queryParams.showThreads === 'true';
+  const showOutputAndEnvVars = queryParams.showOutputAndEnvVars === 'true';
+
+  useEffect(() => {
+    if (location.hash) {
+      localStorage.setItem(
+        SIGN_IN_RESTORE_URL_KEY,
+        `${location.pathname}${location.search}`,
+      );
+      router.replace(SIGN_IN_CALLBACK_PATH + location.hash);
+    }
+  }, [router]);
+
+  return (
+    <AgentRunner
+      namespace={namespace}
+      name={name}
+      version={version}
+      showLoadingPlaceholder={true}
+      showThreads={showThreads}
+      showOutputAndEnvVars={showOutputAndEnvVars}
+    />
+  );
+}
--- a/hub/demo/src/app/sign-in/callback/page.tsx
+++ b/hub/demo/src/app/sign-in/callback/page.tsx
@@ -13,6 +13,7 @@ import { useRouter } from 'next/navigation';
 import { useEffect, useState } from 'react';
 import { type z } from 'zod';
 
+import { useIsEmbeddedIframe } from '~/hooks/embed';
 import {
   extractSignatureFromHashParams,
   MESSAGE,
@@ -33,12 +34,12 @@ export default function SignInCallbackPage() {
     typeof authorizationModel
   > | null>(null);
   const [isInvalidToken, setIsInvalidToken] = useState(false);
+  const { isEmbedded } = useIsEmbeddedIframe();
 
   useEffect(() => {
     async function parseToken() {
       try {
         const hashParams = extractSignatureFromHashParams();
-        const nonce = returnSignInNonce();
 
         const auth = authorizationModel.parse({
           account_id: hashParams?.accountId,
@@ -47,7 +48,7 @@ export default function SignInCallbackPage() {
           callback_url: returnSignInCallbackUrl(),
           message: MESSAGE,
           recipient: RECIPIENT,
-          nonce,
+          nonce: hashParams?.nonce || returnSignInNonce(),
         });
 
         setParsedAuth(auth);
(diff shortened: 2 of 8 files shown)


ewiner:
FYI you've got a build failure here

hub/demo/src/trpc/routers/auth.ts
```diff
@@ -7,11 +7,24 @@ import { createTRPCRouter, publicProcedure } from '../trpc';
...
+  <iframe
+    src="https://app.near.ai/embed/..."
+    sandbox="allow-scripts allow-popups allow-same-origin allow-forms"
+  />
 */
 
 export const AUTH_COOKIE_NAME = 'auth';
 const AUTH_COOKIE_MAX_AGE_SECONDS = 31536000 * 100; // 100 years
-const AUTH_COOKIE_STANDARD_FLAGS = `SameSite=Strict; Path=/; HttpOnly; Secure`;
+const AUTH_COOKIE_STANDARD_FLAGS = `SameSite=None; Path=/; HttpOnly; Secure`;
```
calebjacob, line 27: @ewiner after a bunch of testing and research, the only way to make auth work inside an embedded `<iframe>` on any external domain is by using `SameSite=None` on our auth cookie.

This is one of the better write ups I found that describe pros/cons of `SameSite` configurations: https://andrewlock.net/understanding-samesite-cookies/#samesite-strict-cookies

This article also does a good job outlining how an app can be vulnerable to CSRF attacks: https://engineering.leanix.net/blog/all-you-need-to-know-about-CSRF-attacks/#should-i-be-worried-about-a-csrf-attack

- We don't have CORS enabled for our Next JS app
- We don't have any GET requests that produce side effects

What do you think? Do you have any concerns?

calebjacob: Maybe it'd be a good idea to invest in setting up automated security scans that can detect things like CSRF vulnerabilities.

ewiner: Those things are the worrrrssssttttt.  Though maybe the false positive rate is lower with LLM analysis of the surrounding codebase.

calebjacob: Right, I'd imagine there are some pretty bad security tools out there. Maybe after we do our big rework of auth to support social sign in and whatnot we could consider a security audit of sorts (automated or manual).

calebjacob:
@ewiner @thisisjoshford I pushed a couple small changes that should help improve the white label UX when embedding the iframe. You can optionally define a `details.agent.embed.logo: "https://image.com/foobar.png"` key in your `metadata.json` to render a logo in the main nav.

I also removed the main footer in favor of a `Powered by NEAR AI` logo in the header. This is a link that will open the agent details page in a new tab. EG: https://app.near.ai/agents/gagdiez.near/docs-gpt/latest

Screenshot 2025-02-25 at 4 04 17 PM

If the embed logo isn't provided, we'll just render the `name` of the agent as the logo:

Screenshot 2025-02-25 at 4 09 11 PM

hub/demo/src/trpc/routers/auth.ts
ewiner: Had a think about this... `SameSite=None` means any site could make a `fetch()` to a POST endpoint on app.near.ai and it would be authenticated as the user.  CORS would prevent them from reading the response but the request would still make it through.

calebjacob:
@ewiner @thisisjoshford Just pushed some more changes that will allow users to easily access the embed code for an agent via our agent share button: 

Screenshot 2025-02-25 at 5 24 32 PM

Screenshot 2025-02-25 at 5 24 41 PM

Screenshot 2025-02-25 at 5 58 13 PM

hub/demo/src/trpc/routers/auth.ts
calebjacob: @ewiner I learned something new today - POST requests do go through to the server even though the browser throws an error.

Went through further testing and it appears that even though the POST does go through, the `cookies` are not present when attempting a cross origin POST: https://www.loom.com/share/b655865cd439491cad02a7e6733616cd

Feel free to dig further and let me know what you think!

calebjacob: @ewiner So I figured out where the POST vulnerability would potentially arise. Say you have a malicious website with a vanilla HTML form using `method="POST"`:

```html
<form action="https://drum-emerging-similarly.ngrok-free.app/api/test" method="POST">
  <button type="submit">Submit</button>
</form>
```

This would redirect to the `action` URL and trigger a working `POST` request with all cookies present. This sounds like a deal breaker, but our Next JS codebase would actually prevent this from working thanks to how we use `tRPC` for all API interactions.

A native HTML form submit is only capable of sending a POST request with `ContentType: application/x-www-form-urlencoded`. Our tRPC middleware only respects requests with `ContentType: application/json` and a specific, nested structure.

From what I'm seeing, it should be impossible for an HTML form request to pull this off.

calebjacob: Further digging on `<form>` risks:

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#enctype
- https://www.w3.org/TR/html-json-forms (appears to be dead)

From the W3 proposal:
> 5. Form Submission ... Given that there exist deployed services using JSON and ambient authentication, and given that form requests are not protected by the same-origin policy by default, if this encoding were left wide open then a number of attacks would become possible. Because of this, when using the application/json form encoding the same-origin policy is enforced.

ewiner: Alright, I'm sold!

PR #947, opened 2025-02-25
Adding Datadog extension to runner Dockerfile

andregustavo-plank:
# Description
- Adding Datadog Extension to Runner Lambda Dockerfile

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] These changes may affect the database(data or structure)
- [x] Improving Observability/Traceability 

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Has unit tests
- [ ] Has integration tests
- [x] Only smoke tests

## Action to take when issues are discovered

- [x] Revert PR
- [ ] Disable the feature toggle(Describe it)
- [ ] Other (please specify)

### DIFF
--- a/aws_runner/py_runner/Dockerfile
+++ b/aws_runner/py_runner/Dockerfile
@@ -14,9 +14,7 @@ RUN pip install --upgrade pip
 # Add Datadog Lambda Extension
 COPY --from=public.ecr.aws/datadog/lambda-extension:latest /opt/extensions/ /opt/extensions/
 
-RUN pip install \
-    datadog-lambda \
-    ddtrace
+RUN pip install datadog-lambda
 
 ENV DD_TRACE_ENABLED=true \
     DD_MERGE_XRAY_TRACES=true \
--- a/aws_runner/py_runner/Dockerfile
+++ b/aws_runner/py_runner/Dockerfile
@@ -20,9 +20,7 @@ RUN pip install \
 
 ENV DD_TRACE_ENABLED=true \
     DD_MERGE_XRAY_TRACES=true \
-    DD_LOG_LEVEL=debug \
     DD_LAMBDA_HANDLER="nearai/aws_runner/service.handler" \
-    DD_TRACE_DEBUG=true \
     PYTHONPATH="/opt/python:${PYTHONPATH}" \
     DD_ENHANCED_METRICS=true
 
--- a/aws_runner/py_runner/Dockerfile
+++ b/aws_runner/py_runner/Dockerfile
@@ -10,6 +10,23 @@ ARG FRAMEWORK=""
 COPY aws_runner/frameworks/requirements${FRAMEWORK}.txt ${LAMBDA_TASK_ROOT}
 
 RUN pip install --upgrade pip
+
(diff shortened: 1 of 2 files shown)


docker-compose.yaml
```diff
@@ -14,6 +14,8 @@ services:
       - "9009:8080"
     volumes:
       - ~/.nearai/registry:/root/.nearai/registry
```
zavodil, line 16: Why do you mount local registry to the AWS Instance???

zavodil (changes requested):

docker-compose.yaml
andregustavo-plank: This docker-compose serves only to run locally, this volume was needed to run the lambda runner locally. It will not be used in any AWS instance

andregustavo-plank: https://github.com/nearai/nearai/blob/5c24ce2f6c7c891a3c894b58807d921fc8f46eec/aws_runner/README.md#L65

it's described in this doc...

ewiner (approved):
lgtm

aws_runner/py_runner/Dockerfile
```diff
@@ -10,6 +10,21 @@ ARG FRAMEWORK=""
 COPY aws_runner/frameworks/requirements${FRAMEWORK}.txt ${LAMBDA_TASK_ROOT}
 
 RUN pip install --upgrade pip
+
+# Add Datadog Lambda Extension
+COPY --from=public.ecr.aws/datadog/lambda-extension:latest /opt/extensions/ /opt/extensions/
+
+RUN pip install \
+    datadog-lambda \
+    ddtrace
```
ewiner, line 19: `ddtrace` is already in the `requirements.txt` file of all the runners.  Do we need it here too, where we might end up installing the wrong version?

docker-compose.yaml
zavodil: @andregustavo-plank In the documentation you provided, there are different instructions for setting up local runners. Is there a guide on how to use this specific runner?

Also, I don't understand why `DD_TRACE_ENABLED=true` is set by default in local runners. It seems like it should be disabled because, first, a developer can easily see all their data locally, and second, local files are much more vulnerable to being sent to external services.

andregustavo-plank: This runs the same runner in the local environment, just using docker-compose, this way is easier to pass env vars without having to write long command lines. 

We already have the flag for not enabling the Datadog traces unless we have the DD_API_KEY

andregustavo-plank: Anyway, I updated it all to false.

aws_runner/py_runner/Dockerfile
andregustavo-plank: you are right I updated it. thanks.

zavodil (approved):

PR #953, opened 2025-02-26
Fix iframe embed logo in safari, add support for more embed options

calebjacob:
- Fixed embed logo stretching bug for mobile and Safari
- More agent metadata customization options
- Set `agent.embed.logo: false` to display no logo or text in header
- Set `agent.assistant_role_label` to override default `Assistant` label on each message from the `assistant` role
- Allow developer to set theme via query param. This would allow them to dynamically change the theme between light/dark mode based on their own website user preferences.

Screenshot 2025-02-26 at 11 53 14 AM


thisisjoshford (approved):
Looks good, tysm!

PR #954, opened 2025-02-26
Feat/standalone scheduler

zavodil:
feat: separate the scheduler into its own standalone process that can be run independently from the API server. This approach provides several benefits:
- Reduces potential resource leakage between processes
- Allows the API to run with multiple workers without scheduler duplication
- Simplifies monitoring and debugging of each component

feat: provide Fastnear api key to be set as bearer
fear: env.reply to duplicate message in stdout
fix: local composer default timezone set to UTC


zavodil:
this PR was is currently on staging (I've tested start, restart, logs)

setup_and_run.sh
```diff
@@ -0,0 +1,238 @@
...
+            fi
+        else
+            log "API process (PID: $api_pid) not running"
+        fi
+        rm api.pid
+    fi
+
+    # If the API process is still running (PID didn't work or file wasn't there), kill by name
+    pkill -f "nearai-hub-worker" || true
+    pkill -f "multiprocessing.spawn" || true
```
ewiner, line 181: At this point it feel like we're reimplementing `systemd` from scratch.  WDYT about moving over to `systemd`?  We could keep this script but it would get a lot simpler, just pip install and `systemctl service restart ...`.

setup_and_run.sh
```diff
@@ -0,0 +1,238 @@
...
+    echo "$(date '+%Y-%m-%d %H:%M:%S') - $1"
+}
+# Error handling
+set -e
+trap 'log "Error occurred at line $LINENO. Command: $BASH_COMMAND"' ERR
+# Step 1: Make sure we're in the correct directory
+check_directory() {
+    local path=$1
+
+    if [[ ! -d $path ]]; then
```
ewiner, line 16: If the directory doesn't exist, then we `cd` to it?  I think this is backwards from what you were trying to do.

hub/scheduler.py
```diff
@@ -0,0 +1,204 @@
...
+from hub.tasks.x_event_source import x_events_task
+
+# Configure logging
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+)
+logger = logging.getLogger("nearai-scheduler")
+
+load_dotenv()
```
ewiner, line 30: Please set up `ddtrace` like we do in `app.py`.

hub/scheduler.py
```diff
@@ -0,0 +1,204 @@
...
+    if scheduler.running:
+        logger.info("Shutting down scheduler...")
+        scheduler.shutdown()
+
+    tasks = [t for t in asyncio.all_tasks() if t is not asyncio.current_task()]
+
+    if tasks:
+        logger.info(f"Cancelling {len(tasks)} outstanding tasks")
+        for task in tasks:
+            task.cancel()
```
ewiner, line 190: Should we wait for `task.done()` so we don't fail any tasks in progress?

ewiner (changes requested):
Good changes to make, just don't want to reinvent all the wheels.

setup_and_run.sh
zavodil: Yes, agree

setup_and_run.sh
zavodil: `hub service` pr was existed some time ago https://github.com/nearai/nearai/pull/117/files we can reopen it

this script was created mostly to setup a hub after the update

hub/scheduler.py
zavodil: > Using task.done() checks would create race conditions - they only report status but don't wait for completion. gather() actively awaits termination, ensuring all tasks finish cleanup (even if they ignore CancelledError). Your current code is safer.

arktoswb (changes requested):

hub/scheduler.py
```diff
@@ -0,0 +1,233 @@
...
+                        },
+                        tool_resources=None,
+                        owner_id=auth_token.account_id,
+                    )
+
+                    if task.thread_id is None:
+                        thread = _create_thread(thread_model, auth=auth_token)
+                        task.thread_id = thread.id
+
+                    model = task.run_params.get("model", DEFAULT_MODEL)
```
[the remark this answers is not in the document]
zavodil: Great catch, thank you for review!

nearai/agents/environment.py
```diff
@@ -602,6 +602,9 @@ def add_reply(
             thread_id: str = self._thread_id,
         ):
             """Assistant adds a message to the environment."""
+            if message_type is None or not message_type.startswith("system:"):
+                print(f"add_reply: {message}")
```
arktoswb, line 606: Remove?

nearai/agents/environment.py
```diff
@@ -630,6 +633,8 @@ def _add_message(
             attachments: Optional[Iterable[Attachment]] = None,
             **kwargs: Any,
         ):
+            print(f"add_message [{role}]: {message}")
```
arktoswb, line 636: Remove?

arktoswb (approved):

hub/scheduler.py
ewiner: We would still keep the `gather()` and wait for completion.  `task.done(0` wasn't right – but ideally we could give in-progress tasks a few seconds to complete gracefully before hitting the `cancel()`.

ewiner (approved):
alright let's get this out, but i think we'll need some additional changes afterwards for process management, tracing, etc

nearai/agents/environment.py
zavodil: No it's intentional, it's hard to check from logs what happens otherwise.

nearai/agents/environment.py
zavodil: No it's intentional, it's hard to check from logs what happens otherwise.

arktoswb: It's spamming `nearai agent interactive`

PR #958, opened 2025-02-27
Add necessary steps to run local hub

gagdiez:
I was going through using the Hub local and found out that some steps are missing, adding them on this PR


thisisjoshford (approved):
;)

PR #959, opened 2025-02-27
Modified Message content to Union[List, str]

gagdiez:
So far we are always casting all messages to strings, which destroys the structure of the `messages`

While most of the time this is not a problem, with models that are expecting a very specific structure (e.g. `phi-3-vision-128k-instruct`) passing a string instead of a `List[dict]` will make the model hallucinate and actually fail at the task they are asked to perform

This PR modifies the behaviour of our `ensure_string_content` function, so it iteratively transforms the `values` within the list passed to `messages` instead of the list itself

### DIFF
--- a/hub/api/v1/completions.py
+++ b/hub/api/v1/completions.py
@@ -1,7 +1,7 @@
 import json
 from enum import Enum
 from os import getenv
-from typing import Callable, List
+from typing import Callable, List, Union
 
 from dotenv import load_dotenv
 from nearai.shared.client_config import DEFAULT_MAX_RETRIES, DEFAULT_TIMEOUT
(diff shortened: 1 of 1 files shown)


hub/api/v1/completions.py
```diff
@@ -62,12 +62,19 @@ class Message(BaseModel):
     """A chat message."""
 
     role: str
-    content: str = ""
+    content: List[dict]
```
arktoswb, line 65: https://platform.openai.com/docs/api-reference/messages/createMessage

That should be string or array

gagdiez: modified to be a Union[List, str]

arktoswb (approved):

gagdiez:
@arktoswb I like the idea of going iteratively. 

Let me add a bit of context here to why this is needed, and thus what we might need to change. To [interact with visual models](https://docs.fireworks.ai/guides/querying-vision-language-models) we need the following logic:

```py
# Load image
with open("image.jpeg", "rb") as image_file:
    image_base64 = base64.b64encode(image_file.read()).decode("utf-8")

messages = [
    {
        "role": "user",
        "content": [
            {
                "image_url": {
                    "url": f"data:image/jpeg;base64,{image_base64}"
                },
                "type": "image_url"
            },
            {
                "text": "what do you see in the image? Tell me about it at length.",
                "type": "text"
            }
        ],
    }
]
```

Instead of allowing `random strings` or `random list of strings`, we might want to add classes for the types of messages we accept:

```
User
  - role: str = "user"
  - content: Union[Text, ImageUrl]

System:
  - role: str = "system" 
  - content: Union[Text, Image]
```

In the terminal interactive session we only need to display "Image" whenever in a thread an image is being used 

For the `hub`, we will need to add a button to add images, and the ability to display them

PR #960, opened 2025-02-27
fix: add all folders in the root directory as potential modules to im…

zavodil:
…port


ewiner (approved):
seems reasonable and minimal

PR #961, opened 2025-02-28
Enable multiline input in nearai agent interactive

arktoswb:
Tested with `Enter`s and copy-paste


nearai/cli.py
```diff
@@ -577,7 +577,11 @@ def interactive(
 
         last_message_id = None
         print(f"\n=== Starting interactive session with agent: {agent_id} ===")
-        print("Type 'exit' to end the session\n")
+        print("")
+        print("Type 'exit' to end the session")
+        print("On Linux/macOS: Press Ctrl+D at the beginning of a new line after the prompt")
```
PierreLeGuen, line 582: I'm not understanding the guidance here. So by default `Enter` will not send the message but go to new line? You've to press Ctrl + D to send the msg?

arktoswb: Yes, press Ctrl+D on Linux/Mac at the beginning of a new line to send the msg. That's EOF symbol.

thisisjoshford, line 582:
```suggestion
        print("On Linux/macOS: To submit, press Ctrl+D at the beginning of a new line after your prompt")
        print("Type 'exit' to end the session")
```

Based on the above conversation, I would phrase it this way. Would it be possible to just have it submit by pressing Ctrl+D at any time rather on a new line & at the beginning?

thisisjoshford (approved):
Left a minor suggestion :)

nearai/cli.py
arktoswb: Sending EOF is OS-specific. I don't think I can change it.

PR #966, opened 2025-02-28
Support .gitignore file when uploading to registry

### LINKED ISSUE #860: Support .gitignore when uploading agent files

arktoswb:
Also printing list of files to be uploaded:

```
[0] nearai registry upload /Users/antonlomonos/.nearai/registry/alomonos.near/gitignore-test/latest
.gitignore file detected. Will filter out git ignore files.

47155 files are filtered out and will not be uploaded.

Files to be uploaded:
U   agent.py

100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 296/296 [00:01<00:00, 274B/s]
namespace='alomonos.near' name='gitignore-test' version='0.0.2'
```

Fixes #860

### DIFF
--- a/nearai/registry.py
+++ b/nearai/registry.py
@@ -230,6 +230,7 @@ class Registry:
             gitignore_path = path / ".gitignore"
             if gitignore_path.exists() and gitignore_path.is_file():
                 with open(gitignore_path, "r") as f:
+                    print(".gitignore file detected. Will filter out git ignore files.\n")
                     # Start with Git's default ignore patterns
                     default_ignore_patterns = [
                         # Git internal directories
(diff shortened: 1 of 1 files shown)


PierreLeGuen:
Would be nice to print to the user which files are being ignored from his folder - might save some headaches.

arktoswb:
> Would be nice to print to the user which files are being ignored from his folder - might save some headaches.

The way it's implemented right now, it's not trivial. In the example I used for test there is one file to be uploaded while hundreds of files to be not uploaded. The filtering is done against individual files and not folders, so to make a message about which root level folders are ignored is not trivial.

PierreLeGuen:
@arktoswb what about saying x number of files were ignored?

arktoswb:
> @arktoswb what about saying x number of files were ignored?

Done

PierreLeGuen (approved):

thisisjoshford (commented):
Awesome, ty!

PR #968, opened 2025-02-28
feat: add objects as options for combobox in request_data country-name

bernardomafra-plank:
Started working on that
-  I think its pretty much done for countryName, but I've still couldn't test it right, maybe I'm missing on something
- Missing implementation for 2 letter format in province


hub/demo/src/app/api/v1/aitp/data/forms/shipping_address_international.json/route.ts
```diff
@@ -73,258 +74,258 @@ export async function GET() {
         type: 'combobox',
         autocomplete: 'country-name',
```
calebjacob, line 75: I believe we want to switch `autcomplete: 'country-name'` to just `'country'`: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#country

calebjacob:
Looks good! Were you wanting to keep this as a draft for a bit longer?

bernardomafra-plank:
> Looks good! Were you wanting to keep this as a draft for a bit longer?

I'm still testing it, its not working well yet but it should be something simple

calebjacob (approved):
Hey @bernardomafra-plank, I pushed a bug fix to the `@near-pagoda/ui` combobox component. It was related to having `allowCustomInput: true` and `options: {label: string; value: string;}[]` at the same time.

That bug you came across has been fixed, but I also pushed a change to switch our shipping forms to use `select` instead of `combobox` for country and state. The `select` input type forces the user to select an option whereas the `combobox` input type allows to select an option or enter a custom value. Custom input values for country/state wouldn't make much sense.

bernardomafra-plank:
> Hey @bernardomafra-plank, I pushed a bug fix to the `@near-pagoda/ui` combobox component. It was related to having `allowCustomInput: true` and `options: {label: string; value: string;}[]` at the same time.
> 
> That bug you came across has been fixed, but I also pushed a change to switch our shipping forms to use `select` instead of `combobox` for country and state. The `select` input type forces the user to select an option whereas the `combobox` input type allows to select an option or enter a custom value. Custom input values for country/state wouldn't make much sense.

@calebjacob thanks for that! It is working well now. I just pushed another change to use "select" in the province/state on the shipping address international form as well - also added 2 letter code there - Let me know if that is a problem.

calebjacob:
@bernardomafra-plank Unfortunately, we can't make the `international` province field a select since there's way too many options per country. We'll need to leave that as a text input for now.

You can definitely apply the select options to the `us` state field though.

bernardomafra-plank:
> @bernardomafra-plank Unfortunately, we can't make the `international` province field a select since there's way too many options per country. We'll need to leave that as a text input for now.
> 
> You can definitely apply the select options to the `us` state field though.

Makes sense! I just changed it

calebjacob (approved):

PR #973, opened 2025-03-03
feat: add auto versioning

### LINKED ISSUE #916: Streamline agent version management in CLI for code publishing
## Description
Currently, publishing new agent versions requires manual steps, including creating new folders, renaming directories to match version numbers (e.g., 0.0.2), updating `metadata.json`, and _then_ re-uploading content. This process is error-prone and time-consuming. Would enhance DevX to have a more automated and elegant solution for version management.

## Current Workflow
1. Manually create a new folder
2. Manually rename folder to new version number (e.g., 0.0.2)
3. Update metadata.json file with new version information
4. Re-upload all content
5. Repeat for each version update

## Proposed Solution
Add CLI commands to handle version management:

```bash
# Example commands
nearai agent version bump <AGENT> [major|minor|patch]
or
nearai agent registry upload <AGENT> --auto-increment
```

thisisjoshford:
fixes #916 

- Checks for existing version in `metadata.json` and if exists throws error prompting user to update w/ options
- Adds `--bump`, `--bump-minor`, and `--bump-major` flags to upload that auto increments `metadata.json`
- Adds alias for `agent upload` that uses  `registry update`
- Ensures PEP 440 versioning is used and throws error if not
- Adds registry tests for TDD :) 

---

Example error w/ suggestions:

Screenshot 2025-03-09 at 10 57 01 PM

Version Update example:

Screenshot 2025-03-09 at 10 57 43 PM


zavodil:
@thisisjoshford If in this PR you will be performing a check of the agent version from the provided metadata, please ensure that the version is validated to be compatible with the Python Packaging tooling: PEP 440 Versioning. https://packaging.pypa.io/en/stable/version.html
We cannot accept the version as just any string without a format, at least to enable proper version comparison in the Discovery Agent.

thisisjoshford:
> @thisisjoshford If in this PR you will be performing a check of the agent version from the provided metadata, please ensure that the version is validated to be compatible with the Python Packaging tooling: PEP 440 Versioning. https://packaging.pypa.io/en/stable/version.html We cannot accept the version as just any string without a format, at least to enable proper version comparison in the Discovery Agent.

Thanks for this callout @zavodil !

thisisjoshford:
@zavodil To ensure users conform to PEP 440 I've used `Version` from `.venv/lib/python3.11/site-packages/packaging/version.py` to create this simple  [validate_version](https://github.com/nearai/nearai/pull/973/files#diff-48fd8840b47760564a04b4bf16f7021ea5257d2e6521b87fd5524a07ab7a1cd5R80-R96) method. 

As I was writing tests for the features in this PR I found some strange behavior when creating tests for ensuring PEP 440 was being followed. ([See commented out code](https://github.com/nearai/nearai/pull/973/files#diff-2349d7c101e3fd33808218570c6cd5f8682e71e3bbb30f62d618340a886c1d94R324-R347)) Should we put more restrictions on checking versioning than what `version.py` provides w/ python 3.11?

nearai/cli.py
```diff
@@ -174,34 +190,100 @@ def list(
                     f"There are unregistered common provider models: {unregistered_common_provider_models}. Run 'nearai registry upload-unregistered-common-provider-models' to update registry."  # noqa: E501
                 )
 
-    def update(self, local_path: str = ".") -> None:
-        """Update metadata of a registry item."""
-        path = resolve_local_path(Path(local_path))
+    def update(self, local_path: str = ".", increment_type: str = "patch") -> None:
+        """Update the version in metadata.json file.
```
arktoswb, line 194: This function can update any field in metadata.json, not just the version.

nearai/cli.py
```diff
@@ -174,34 +190,100 @@ def list(
...
 
-        if CONFIG.auth is None:
```
arktoswb, line 181: Why was it removed?

nearai/cli.py
```diff
@@ -174,34 +190,100 @@ def list(
...
+
+        # Get current version
+        current_version = metadata["version"]
 
-        entry_location = EntryLocation.model_validate(
-            dict(
-                namespace=namespace,
-                name=metadata.pop("name"),
-                version=metadata.pop("version"),
+        # Increment version based on increment_type
```
arktoswb, line 221: There seems to be a lot of code duplication between `def update` and `def update_agent_version`

nearai/cli.py
thisisjoshford: ah, I thought this was redundant to check login here as I noticed we are not checking auth status on other commands performed (upload for example):

https://github.com/nearai/nearai/blob/18d4367b1354aa2ebb538121b5bd10097dda2e8a/nearai/cli.py#L257-L260

I'll add it back and perhaps we should add some auth checks after other commands as well.

nearai/cli.py
thisisjoshford: ah, nice catch. Let me resolve this.

nearai/cli.py
thisisjoshford: Good call, let me see if I can refactor this a bit.

thisisjoshford: agent update is only supposed to be an alias for registry update and had some duplicate logic in a merge from main conflict resolve. Updated now :)

nearai/cli.py
thisisjoshford: Updated on L271-272

arktoswb: I meant, the method was initially designed to update any field in metadata.json. You have changed it to only update the version. The reasoning behind this change is not clear to me.

thisisjoshford: ah, I see now. When first starting this PR to fix #916 I was under the assumption this method bumped version ([based on this comment in the issue](https://github.com/nearai/nearai/issues/916#issuecomment-2691786889)) and was wanting to add a simple way to bump version using `--minor` / `--major` flag with default being `patch`. Now re-reading the comment and looking closer at this method, I see its intended purpose is for updating the remote registry's `metadata.json` with the contents of local `metadata.json` and not bumping a local version before upload. Thanks for the clarification here. 

Let me rethink my approach here. Perhaps I should just leave this method as is and just leave the modification of `upload` to allow for `--auto-increment` with optional `--minor` / `--major`

nearai/cli.py
```diff
@@ -254,9 +272,124 @@ def upload_unregistered_common_provider_models(self, dry_run: bool = True) -> No
...
+        assert metadata is not None, "Metadata should not be None if error is None"
+
+        name = metadata["name"]
+        version = metadata["version"]
+
+        # Get namespace from auth
+        if CONFIG.auth is None or CONFIG.auth.namespace is None:
+            print("Please login with `nearai login` before uploading")
+            return None
+
```
arktoswb, line 311: Users may upload entries on behalf of orgs. Use get_namespace() function instead.

nearai/cli.py
```diff
@@ -254,9 +272,124 @@ def upload_unregistered_common_provider_models(self, dry_run: bool = True) -> No
...
+                    console.print(update_panel)
+
+                    # Update metadata.json with new version
+                    metadata["version"] = version
+                    with open(metadata_path, "w") as f:
+                        json.dump(metadata, f, indent=2)
+
+                    console.print(f"[dim]Updated [bold]{metadata_path}[/bold] with new version[/dim]")
+                    continue  # Try again with new version
+                else:
```
arktoswb, line 366: Hard to see which `if` this `else` refers to. Maybe refactor to:
```
bump_requested = bump or minor_bump or major_bump
if exists and bump_requested:
  # Do bump
  continue
if exists and not bump_requested:
  # Print error panel
  continue

# Version doesn't exist, proceed with upload
```

arktoswb (changes requested):

nearai/cli.py
thisisjoshford: updated @arktoswb , lmk if this refactoring works for you.

nearai/cli_helpers.py
```diff
@@ -56,6 +61,133 @@ def display_agents_in_columns(agents: list[Path]) -> None:
     console.print("\n")
 
 
+def validate_version(version: str) -> Tuple[bool, Optional[str]]:
```
arktoswb, line 64: Maybe move this method to `registry.py`?

nearai/cli_helpers.py
```diff
@@ -56,6 +61,133 @@ def display_agents_in_columns(agents: list[Path]) -> None:
...
+
+    """
+    try:
+        Version(version)
+        return True, None
+    except InvalidVersion as e:
+        return False, f"Invalid version format: {str(e)}. Version must follow PEP 440:https://peps.python.org/pep-0440."
+
+
+def increment_version_by_type(version: str, increment_type: str) -> str:
```
arktoswb, line 83: Maybe move this method to `registry.py`?

nearai/cli_helpers.py
```diff
@@ -56,6 +61,133 @@ def display_agents_in_columns(agents: list[Path]) -> None:
...
+        return metadata, None
+    except FileNotFoundError:
+        return None, f"Metadata file not found at {metadata_path}"
+    except json.JSONDecodeError:
+        return None, f"Invalid JSON in metadata file at {metadata_path}"
+    except Exception as e:
+        return None, f"Error reading metadata file: {str(e)}"
+
+
+def check_version_exists(namespace: str, name: str, version: str) -> Tuple[bool, Optional[str]]:
```
arktoswb, line 149: Maybe move this method to `registry.py`?

arktoswb (approved):

arktoswb:
> @arktoswb @thisisjoshford this pr was breaking the prod runner because of missing `rich` dependency #1008
> 
> The best fix is to remove rich from https://github.com/nearai/nearai/pull/973/files#diff-6f78f41e937a66d942fef1ad9647818eef49a3fd67f32e0a3830b6561f4a1d9dR7-R8 to avoid adding it into the every runner

Oh my. Maybe we should progress toward separating repos into `nearai-core`, `nearai-cli`, `nearai-hub`, and `nearai-runner`

arktoswb:
> > @arktoswb @thisisjoshford this pr was breaking the prod runner because of missing `rich` dependency #1008
> > The best fix is to remove rich from https://github.com/nearai/nearai/pull/973/files#diff-6f78f41e937a66d942fef1ad9647818eef49a3fd67f32e0a3830b6561f4a1d9dR7-R8 to avoid adding it into the every runner
> 
> Oh my. Maybe we should progress toward separating repos into `nearai-core`, `nearai-cli`, `nearai-hub`, and `nearai-runner`

Actually, the problem will be solved with TEE runners, as TEE runners are in separate repo, right?

thisisjoshford:
> @arktoswb @thisisjoshford this pr was breaking the prod runner because of missing `rich` dependency #1008
> 
> The best fix is to remove rich from https://github.com/nearai/nearai/pull/973/files#diff-6f78f41e937a66d942fef1ad9647818eef49a3fd67f32e0a3830b6561f4a1d9dR7-R8 to avoid adding it into the every runner

https://github.com/nearai/nearai/pull/1013 fixed this

thisisjoshford:
> > > @arktoswb @thisisjoshford this pr was breaking the prod runner because of missing `rich` dependency #1008
> > > The best fix is to remove rich from https://github.com/nearai/nearai/pull/973/files#diff-6f78f41e937a66d942fef1ad9647818eef49a3fd67f32e0a3830b6561f4a1d9dR7-R8 to avoid adding it into the every runner
> > 
> > 
> > Oh my. Maybe we should progress toward separating repos into `nearai-core`, `nearai-cli`, `nearai-hub`, and `nearai-runner`
> 
> Actually, the problem will be solved with TEE runners, as TEE runners are in separate repo, right?

@PierreLeGuen can you confirm this?

PR #974, opened 2025-03-03
feat: setup_and_run.sh to back up old log files before starting new processes

zavodil:
feat: setup_and_run.sh to back up old log files before starting new processes

Moved existing api.log and scheduler.log to timestamped backups before starting new logs.

### DIFF
--- a/setup_and_run.sh
+++ b/setup_and_run.sh
@@ -56,6 +56,9 @@ run_api() {
 
     log "Starting FastAPI server with $workers workers on port $port..."
 
+    # Backup old log file with timestamp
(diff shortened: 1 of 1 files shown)


setup_and_run.sh
```diff
@@ -69,15 +72,20 @@ run_api() {
...
+
     local scheduler_pid=$!
     echo $scheduler_pid > scheduler.pid
     log "Scheduler started with PID $scheduler_pid"
 }
 # Main function to run everything
 main() {
     local workers=${1:-8}
-    local port=${2:-8081}
+    local port=${2:-8001}
```
arktoswb, line 88: Why the port number change?

zavodil: our default port for prod is 8001

arktoswb (approved):

PR #858, opened 2025-02-14
Supporting running benchmark on url

ilblackdragon:
Example:
```
nearai benchmark run "near.ai/mbpp/1.0.0" MBPPSolverStrategy --model https://api.openai.com/v1::gpt-4o  --subset=train --max_concurrent=1
```

If the url requires API key, can specify it via:
```
export OPENAI_API_KEY = "<key>"
```

PR #1000, opened 2025-03-07
Fix url provider bug in evaluation entry

arktoswb:
Follow up to https://github.com/nearai/nearai/pull/858

Url providers like 'https://api.openai.com/v1' can't be included into registry entry name because of special characters.

### DIFF
--- a/nearai/evaluation.py
+++ b/nearai/evaluation.py
@@ -1,6 +1,6 @@
 import json
-from pathlib import Path
 import re
+from pathlib import Path
 from textwrap import fill
 from typing import Any, Dict, List, Union
 
--- a/nearai/evaluation.py
+++ b/nearai/evaluation.py
@@ -109,6 +109,8 @@ def upload_evaluation(
         key += f"_version_{version}"
     if provider != "":
         metrics[EVALUATED_ENTRY_METADATA]["provider"] = provider
+        # Url providers like 'https://api.openai.com/v1' can't be included into registry entry name
+        # because of special characters.
         clean_provider = re.sub(r'[^a-zA-Z0-9_\-.]', '_', provider)
         key += f"_provider_{clean_provider}"
 
--- a/nearai/evaluation.py
+++ b/nearai/evaluation.py
@@ -111,7 +111,7 @@ def upload_evaluation(
         metrics[EVALUATED_ENTRY_METADATA]["provider"] = provider
         # Url providers like 'https://api.openai.com/v1' can't be included into registry entry name
         # because of special characters.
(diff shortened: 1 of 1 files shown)


ewiner:
What's the purpose of having a "URL provider" in the first place?

arktoswb:
> What's the purpose of having a "URL provider" in the first place?

So that benchmarks can be run against a proxy. Cogniscient is currently running NEAR AI benchmarks against their model using a url proxy.

Example command: `nearai benchmark run alomonos.near/mbpp_100samples/0.0.1 MBPPSolverStrategy --model https://api.openai.com/v1::gpt-4o  --subset=test  --max_concurrent=1`

I have added Chat-gpt models into evaluation table.

ewiner:
I see.  What I mean is, why is the OpenAI provider called `https://api.openai.com/v1` instead of `openai`?

arktoswb:
> I see. What I mean is, why is the OpenAI provider called `https://api.openai.com/v1` instead of `openai`?

Semantically-wise, it does not seem to be important.
Implementation-wise, because benchmark called this url directly instead of calling provider through our hub.

arktoswb:
The provider field can be updated manually in evaluation entries any time.

arktoswb:
We can't really determine provider on the fly from user-owned proxies like `http://192.168.1.1/v1`, so the options are:
1. Record `http://192.168.1.1/v1` as provider, with an option to update provider later. Which is what this PR is doing.
2. Or ask for a provider name explicitly during evaluation entry creation.

ewiner:
I guess my issue here is more with #858.  Two matters to tackle:
* execution: making sure it works remotely if possible or fails clearly if impossible
* evaluation: making sure we know what results are verified or verifiable with remote (trusted) execution

For the purposes of execution, the URL would have to be OpenAI-compatible, right?  So that works for evaluating against OpenAI, but if you also want to evaluate against Claude, Gemini, Hosted Deepseek, etc, then you're right back to building new named providers anyway.  Is there much of a use case for URL providers that aren't https://api.openai.com?

The obvious other usage, which you mentioned, is an OpenAI-compatible proxy like LiteLLM.  But:
* if it's hosted / public, might as well make a named provider.
* if it's local / private, why should we be putting the implementation details of your machine's setup in the registry DB?  rather it say something like 'local' merely indicating that this execution is not replicable remotely.

I'm not gonna hold up this PR but I'd like to complete the conversation when you get a chance.

ewiner (approved):

arktoswb:
Yes, I understand the concerns and will investigate further. This PR is a positive improvement (crashes before PR, works with possibility of updates after PR), so I am going to merge it.

PR #1003, opened 2025-03-10
Organization System Implementation

zavodil:
Added Org, OrgMember, and OrgInvitation models with:
- Unique namespace enforcement
- Role-based permissions (admin/maintainer/member)
- Invitation lifecycle management
- Created indexes for optimized queries
- Registry Entry Ownership Expansion

Added owner_type (user/org) and author fields to RegistryEntry
- Introduced composite unique index (namespace, owner_type)
- Implemented create_for_org() factory method with org validation
- Cross-Model Permission Support

Added owner_type field to related models:
- AgentData
- HubSecrets
- Stars Enabled ownership tracking across core entities
Access Control Logic

Created org.py with check_org_access() function:
- Validates direct org membership
- Checks indirect access via entry authorship
- Backward Compatibility

Maintained original namespace field behavior
- Default owner_type="user" for existing entries
- Set author=None for non-org entries

Key Technical Decisions:

- Used SQLModel's Enum for type safety
- Implemented permission inheritance pattern
- Added index-optimized query paths
- Maintained existing API contracts
- Separated org logic into dedicated module

Migration-Ready Features:

- Org/User namespace coexistence
- Granular permission management
- Org member lifecycle tracking
- Cross-model ownership tracking


arktoswb:
What will be the naming format? `*.near` for individuals and `*.ai` for organizations?

zavodil:
> What will be the naming format? *.near for individuals and *.ai for organizations?

I'm not sure if we are in position to force this. I thought just about any NEAR account, but not a `just-a-plain-text-username`.

arktoswb:
> Maybe we started development too early

Yes, a development was started too early. There is a final decision on "Auth System Design", but there is no design doc. And I personally would love to see more docs about our database.

> why this couldn’t be a single namespace

`*.near` for both individual accounts and organizations? We already have `near.ai` organization name, so why not give ability  to other organizations to create names like `near.ai`?

zavodil:
@arktoswb most probably every agent will be published under the org, similar to Netlify and Vercel model

PR #1005, opened 2025-03-10
Fix read_solutions bug of handling --verbose flag

arktoswb:


cmrfrd (approved):
better error handling case :+1:

PR #1006, opened 2025-03-10
Account for name extensions in name matching

arktoswb:
So that:
1. model `jones.near/model` matches with fireworks model `jones/model`
2. check for username <-> organization collisions. Will prevent creation of `jones/model` entry if `jones.near/model` exists.

### DIFF
--- a/nearai/shared/naming.py
+++ b/nearai/shared/naming.py
@@ -8,7 +8,7 @@ def get_canonical_name(name: str) -> str:
 
     Applies such transformations:
     1. All letters lowercase.
-    2. Strips '.near' and '.ai' extensions.
+    2. Strips '.near' extensions.
     3. Convert '.' between digits to 'p'.
     4. Convert '<not letter>v<digit>' -> '<not letter><digit>'
     5. Remove all non-alphanumeric characters except between digits.
@@ -22,9 +22,6 @@ def get_canonical_name(name: str) -> str:
     # Strip .near extension if present
     if name.endswith(".near"):
         name = name[:-5]  # Remove last 5 characters ('.near')
-    # Strip .ai extension if present
-    if name.endswith(".ai"):
-        name = name[:-3]  # Remove last 3 characters ('.ai')
     # Convert '.' between digits to 'p'
     name = re.sub(r"(\d)\.(\d)", r"\1p\2", name)
     # Convert '<digit>v<digit>' -> '<digit>_<digit>'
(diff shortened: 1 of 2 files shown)


nearai/shared/naming.py
```diff
@@ -8,16 +8,20 @@ def get_canonical_name(name: str) -> str:
...
-    5. Convert 'metallama' -> 'llama'.
+    6. Convert 'metallama' -> 'llama'.
 
     e.g. "llama-3.1-70b-instruct" -> "llama3p1_70binstruct"
     """
     # Convert to lowercase
     name = name.lower()
+    # Strip .near extension if present
+    if name.endswith(".near"):
+        name = name[:-5]  # Remove last 5 characters ('.near')
```
zavodil, line 24: I don't know the reasoning behind this, but in my opinion, it will only make our work more complicated. 

For example, if there's an account name.ai.near (which is possible) and you strip it into `name.ai`, it will be indistinguishable from the `name.ai` account in the .ai namespace.

arktoswb: Yes, that's the point. I don't want our registry to be cluttered with entries of similar naming. I don't want anyone to push entries to our registry under the name `near.ai.near`. I also don't want anyone to push entries to our registry under the name `near_ai` (which will be preventable if we include similar names check during account creation step)

arktoswb: Confusions will be either way. If we do similar names check or if we do not do similar names check. I only created a function for similar names check. I don't have a strong opinion on where the similar names check should be used. I think, we should use it (confusions before the fact are better than confusions after the fact), but it's not a strong opinion.

To do an automatic name match between websites is more important matter to me. John Doe has a namespace `john_doe.near` on NEAR AI and `john_doe` on fireworks. I want to do an automatic match between models deployed under `john_doe.near` in NEAR AI registry and models deployed under `john_doe` on fireworks. I would also prefer that we don't have collisions when I do such a match (that is model on fireworks only matches with one entry under NEAR AI and not multiple ones). But that preference should not affect our decision on how it will be best to handle names in our database and registry.

arktoswb: We are not renaming. Canonical name is for comparison purposes only and not visible to a user.

Yes, I see technical issues (and maybe legal issues) for the projects I am working on if we allow both "apple" and "apple.near" accounts to exist on NEAR AI website and allow those accounts to belong to different people.

I understand that it is a valued freedom to have a wallet of any name in Web3. However, we are not a digital wallet website. We are AI website where companies will potentially showcase their models and agents. And so, we have to have more restrictions when it comes to naming.

thisisjoshford (approved):
I see your concerns, and don't have a strong opinion on this as it's just for model uploading. This PR should also update docs or have a corresponding one explaining how this new format will work.

PR #1008, opened 2025-03-11
added 'rich' dependency needed by ddtrace

gabehamilton:

PR #1011, opened 2025-03-11
fix: No module named 'rich'

zavodil:
I see an error Exception: Request failed with status code 502: {"errorMessage": "Unable to import module 'datadog_lambda.handler': No module named 'rich'", "errorType": "Runtime.ImportModuleError", "requestId": "", "stackTrace": []} on a local runner. I can add a missing dependency for runners with the rich package, but do we really need it?


andregustavo-plank (approved):

zavodil:
already addressed in https://github.com/nearai/nearai/pull/1008

PR #1021, opened 2025-03-12
Fixes error when running under uv run

gabehamilton:


cmrfrd (approved):

cmrfrd:
very complicated. but very well thought out

PR #1037, opened 2025-03-14
default runner key for CLI with local hub

gabehamilton:


zavodil:
@gabehamilton  What is the reasoning to set it to `key1`?

gabehamilton:
'local-cli' would be better but I was thinking about current devs who may have already copied over the. .env.example

zavodil (changes requested):
I think this PR will make the code more complex and harder to maintain. If RUNNER_API_KEY is not set, the code should handle the missing value explicitly. If the purpose of these keys is not clearly documented, we should improve the documentation instead.

gabehamilton:
> Will empty value `RUNNER_API_KEY` work after your changes? This is one of the places where empty value is used: https://github.com/nearai/nearai_langchain/blob/main/nearai_langchain/local_nearai_chat_model.py#L16

With the latest changes (documentation rather than a default value), that should work or not work in the same manner as before. The agent code doesn't have access to the RUNNER_API_KEY in the docker runners. But it looks like you are expecting some custom set up anyway so I expect that will work the same.

zavodil (approved):

PR #1053, opened 2025-03-18
Add path error processing message for non-local agent interactive runs

arktoswb:


cmrfrd (approved):
We should put a TODO here to remember that the --local flag is hard coded so we should reference this hard coded string if --local is changed at all in the future

thisisjoshford (approved):
Thanks for this quick fix @arktoswb! 

I think a future follow-up for better resolving this issue could be one of the alternative solutions as mentioned in #1061. I think we should either allow users to run agents remotely this way by converting the listed agents to the proper format or have `agent interactive` w/o args default to `--local` w/ instructions on how to remotely run agents.

PR #1060, opened 2025-03-20
Clean model name for livebench benchmark

arktoswb:
Slashes in model name are valid, but the result `live_bench.evaluated_entry_name` is used as a folder name. So, this PR cleans the characters that are not ok for folder names.


cmrfrd (approved):
Small and simple

Issue #1051, opened 2025-03-17
Fix Bugs in NEAR AI CLI Installation Process


SterFonseca:
The CLI installation process has bugs, creating hurdles for developers. These need to be identified and fixed to ensure a smooth setup.

### Tasks:

[ ] Identify and fix CLI installation bugs.
[ ] Improve error messages and troubleshooting steps.

This issue is based on recent NEAR AI developer feedback highlighting key pain points and improvement areas.

sebmi:
After few install , impossible to go further.
This message appear after ~$ nearai login

File "/home/xxxxxxxxxx/nearai-env/lib/python3.12/site-packages/nearai/agents/environment.py", line 31, in <module>
    from openai.types.beta.vector_store import VectorStore
ModuleNotFoundError: No module named 'openai.types.beta.vector_store'

arktoswb:
How can we prevent people from using obsolete versions of `nearai`?

thisisjoshford:
> After few install , impossible to go further. This message appear after ~$ nearai login
> 
> File "/home/xxxxxxxxxx/nearai-env/lib/python3.12/site-packages/nearai/agents/environment.py", line 31, in from openai.types.beta.vector_store import VectorStore ModuleNotFoundError: No module named 'openai.types.beta.vector_store'

@sebmi Looks like you have two issues here: 

1) `nearai` requires python 3.11.x ( `3.12+` will not work)
2) You are using an outdated version of `nearai` latest is `0.1.16`

Try changing your python version to `3.11.x` and then install `nearai`. Let me know if this resolved your issue.

thisisjoshford:
> How can we prevent people from using obsolete versions of `nearai`?

Fair point @arktoswb, we certainly can't prevent anyone from using older versions. However, I think we could add more prominent/urgent callouts to the user to upgrade to the latest version due to the project being in an `alpha` state with continuous features and fixes being shipped. I could work on a proposed solution here for this for you to review. 

Also, one of the main issues that was reported was how long it took to discover that they needed python `3.11` for the CLI to work. It seems people are not reading the README of the project for instructions (which clearly states this) and instead pointed out that docs were not clear on this. 

1) Its collapsed

![Image](https://github.com/user-attachments/assets/2871f121-dc36-4581-9a02-dfbc5eab7484)

2) Not exactly explicit that you need 3.11

![Image](https://github.com/user-attachments/assets/f5eb1783-992f-4fa8-a5f3-0ebdecf24d83)

[...]

PR #1065, opened 2025-03-22
fix: update cli install instructions

thisisjoshford:
Many users at the recent hackathon reported difficulties installing NEAR CLI as it was not apparent they needed Python 3.11 as mentioned in #1051. 

 This PR adds a more prominent callout for the `3.11` requirement as well as suggesting `uv` as they will need this if they intend to contribute to the codebase or docs.

(Also removed the info callout to NEAR AI hub w/ broken link) I don't feel this is necessary to call out.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -7,10 +7,40 @@ NEAR AI CLI allows you to [create and deploy agents](./agents/quickstart.md), [t
 ## Installing NEAR AI CLI
 
 !!! warning
-    Requires Python version **`3.11`** _(Currently does not work with `3.12` or `3.13`)_
+    Requires Python version **`3.9 - 3.11`** _(Currently does not work with `3.12` or `3.13`)_
 
-!!! tip
-     Although not required, this project uses [uv](https://docs.astral.sh/uv/) for Python package management and it is suggested that you use it as well if you would like to make contributions to this project.
+!!! abstract "Python Version Management"
+
+    If you do not have Python, or your version is not compatible, we recommend using a version management tool as they allow you to easily install and switch between python versions. 
+    
+    Here are a few options to choose from:
+    
+    - [uv](https://docs.astral.sh/uv/) - Fast Python package manager _(Preferred tool by `nearai` core contributors)_
+    - [pyenv](https://github.com/pyenv/pyenv) - Simple Python version management tool
+    - [miniconda](https://docs.anaconda.com/miniconda/install/) - Miniature installation of Anaconda Distribution 
+
+    === "uv"
+
+        ```bash
+        uv venv --python 3.11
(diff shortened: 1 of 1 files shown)


docs/cli.md
```diff
@@ -1,17 +1,17 @@
...
 
 ---
 
 ## Installing NEAR AI CLI
 
+!!! warning
+    Requires Python version **`3.11`** _(Currently does not work with `3.12` or `3.13`)_
+
+!!! tip
+     Although not required, this project uses [uv](https://docs.astral.sh/uv/) for Python package management and it is suggested that you use it as well if you would like to make contributions to this project.
```
gagdiez, line 13: is this really necessary?

docs/cli.md
```diff
@@ -23,28 +23,9 @@ The NEAR AI CLI allows you to interact with NEAR AI's services to [create agents
     ``` bash
     git clone git@github.com:nearai/nearai.git
     cd nearai
-    pip install -e .
+    ./install.sh
     ```
 
-??? abstract "Python Version"
```
gagdiez, line 29: maybe instead of removing, we can add a tab for `uv`?

docs/cli.md
```diff
@@ -23,28 +23,9 @@ The NEAR AI CLI allows you to interact with NEAR AI's services to [create agents
     ``` bash
     git clone git@github.com:nearai/nearai.git
     cd nearai
-    pip install -e .
+    ./install.sh
```
gagdiez, line 26: Instead of pointing to `install.sh` (which assumes I use `venv` to handle my environment) I would still prefer we include instructions on how to do change python version for different environment handlers (`conda`, `venv`, `pyenv`, etc)

docs/cli.md
thisisjoshford: Yep, I think 3.9 & 3.10 work as we have CI that tests these versions but I had someone at a hackathon who had 3.9 and was unable to use it until they upgraded to 3.11. This may have been a windows related issue though.... I can update this to be 3.9-3.11 and see if we get any reports on it.

docs/cli.md
thisisjoshford: ah sure, was just simplifying this

docs/cli.md
thisisjoshford: I wanted to keep installation consistent from the README and docs. I brought up changing to `pip install -e .` to @gabehamilton a week or two ago but he recommended sticking with `venv` as we should have devs running this in isolation instead of installing all of our libraries into their global python.

thisisjoshford: Ok, updated this section... Please re-review @gagdiez

gagdiez (approved):

PR #972, opened 2025-03-03
feat: enhance cli help menu

thisisjoshford:
Introduces robust help menus for CLI developers

- Main Menu Help listing all available commands
- Class Help listing all commands for that class w/ options, examples, & docs
- Individual Method Help w/ options, examples, & docs
- Generates content from docstrings of classes/methods for better maintenance

fixes #971 

---

### Main Menu Help
Screenshot 2025-04-09 at 1 28 53 PM

### Class Help
Screenshot 2025-04-09 at 1 04 20 PM

### Individual Command Help
Screenshot 2025-04-09 at 1 04 40 PM

PR #1067, opened 2025-03-24
Add cli command reference doc

bucanero:


docs/cli_command_reference.md
```diff
@@ -0,0 +1,421 @@
...
+1. [Registry Commands](#registry-commands)
+2. [Config Commands](#config-commands)
+3. [Benchmark Commands](#benchmark-commands)
+4. [Evaluation Commands](#evaluation-commands)
+5. [Agent Commands](#agent-commands)
+6. [Hub Commands](#hub-commands)
+7. [Login Commands](#login-commands)
+8. [Logout Command](#logout-command)
+9. [Permission Commands](#permission-commands)
+10. [Miscellaneous Commands](#miscellaneous-commands)
```
thisisjoshford, line 16: Let's organize this a bit better in the order in which a user would first use it, starting with auth (login/logout) and then going into agent commands (likely the next step), etc. 

See these lines in this PR draft and how I've organized commands: https://github.com/nearai/nearai/pull/972/files#diff-fe41befe061a55df2aab205e13bb082a315064becfb2e875edaf154a98b4436eR1696-R1733

docs/cli_command_reference.md
```diff
@@ -0,0 +1,421 @@
...
+```
+
+Example:
+```
+nearai registry list --category=agent --tags=nlp,chatbot --total=10
+```
+
+### `nearai registry update`
+
+Update metadata of a registry item.
```
thisisjoshford, line 64: We should add more information about some of these commands and ideally test all of them out so we know how they work. This one, for example, only updates the _remote_ metadata.json file from information that is found locally.

thisisjoshford (approved):

PR #1068, opened 2025-03-25
Bernardomafra/nea 80 adapt near tool registry to handle mcp format

bernardomafra-plank:
Enabling a MCP tool to be registered with the tool registry

Agent to test this changes: https://app.near.ai/agents/bernardomafra.near/example-mcp-agent/latest/source

### DIFF
--- a/nearai/agents/models/tool_definition.py
+++ b/nearai/agents/models/tool_definition.py
@@ -12,6 +12,6 @@ class MCPTool(BaseModel):
     """The name of the tool."""
     description: str | None = None
     """A human-readable description of the tool."""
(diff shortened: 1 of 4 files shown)


nearai/agents/models/tool_definition.py
```diff
@@ -0,0 +1,13 @@
+from typing import Any
+from pydantic import BaseModel, ConfigDict
+
+class MCPTool(BaseModel):
```
cmrfrd, line 4: this works for now, but for the future we should try to import this definition from the official mcp python sdk. Can we put a TODO comment here above the class so we know to fix this?

PR #1070, opened 2025-03-26
Add Hub API documentation

bucanero:
Hub API docs are auto-generated by the FastAPI framework.

Adding links for accessibility and easy discovery


mkdocs.yaml
```diff
@@ -177,7 +177,9 @@ nav:
     - Quickstart: 'integration/quickstart.md'
   - APIs:
     - Library API: 'api.md'
-    - Web API: 'https://api.near.ai/openapi.json'
+    - Web API Docs: 'https://api.near.ai/redoc'
+    - Interactive API: 'https://api.near.ai/docs'
+    - OpenAPI Specs: 'https://api.near.ai/openapi.json'
```
thisisjoshford, line 182:
```suggestion
    - Hub Standard API: 'https://docs.near.ai/api/hub/standard'
    - Hub Interactive API: 'https://api.near.ai/api/hub/interactive'
```

I would prefer something closer to this, rather than having both `api.near.ai/docs` AND `docs.near.ai/api`

mkdocs.yaml
```diff
@@ -177,7 +177,9 @@ nav:
     - Quickstart: 'integration/quickstart.md'
   - APIs:
     - Library API: 'api.md'
```
thisisjoshford, line 179:
```suggestion
    - Agent SDK API: 'api.md'
```

thisisjoshford:
@bucanero Left a couple of suggestions, but ideally, I think we should create an overview landing page for `docs.near.ai/api` with a short explainer for each API docs page.

jayzalowitz:
heres some information 

https://github.com/petersalomonsen/quickjs-rust-near/blob/main/examples/aiproxy/web/nearai/login.js

https://github.com/INTEARnear/Tear/blob/e09f63c4393bdabd1410b4d8b3d0575a1df421b1/agents/src/near_ai.rs#L170-L206
https://github.com/prakhar728/near-ai-web-starter

We need a getting started here for logging in and calling externally

mkdocs.yaml
bucanero: URL paths for the auto-generated docs can be customized with:

```py
FastAPI(docs_url="/api/hub/interactive", redoc_url="/api/hub/standard")
```

To-do: we need to decide the best paths for these 2 pages.

thisisjoshford (approved):

PR #1077, opened 2025-04-02
Bernardomafra/nea 82 add mcp support on environment

bernardomafra-plank:
Created an abstraction to simplify the connection with MCP servers

Docs: 
https://github.com/nearai/nearai/compare/bernardomafra/nea-82-add-mcp-support-on-environment?expand=1#diff-f064b66d77d78a070c831421c724302e6476e85dd40680e9a6b2a62d159e9c74R42-R63

Example of agent to test the new feature:
[example-mcp-agent](https://app.near.ai/agents/bernardomafra.near/example-mcp-agent/latest)


cmrfrd (approved):
This looks great! thank you!

nearai/shared/near/serializer.py
```diff
@@ -29,7 +29,7 @@ def serialize_field(self, value, field_type):  # noqa: D102
                 pass
             else:
                 assert len(value) == len(field_type)
-                for v, t in zip(value, field_type):
+                for v, t in zip(value, field_type, strict=False):
```
gabehamilton, line 32: @zavodil Does this change seem ok for completion verifications?

gabehamilton:
Is the MCP state expensive enough to create that it is worth saving in agent_data?
agent_data has one complication. It requires a trusted runner which means agents that use it won't work with a local runner (i.e. near cli) with a remote hub.

bernardomafra-plank:
> Is the MCP state expensive enough to create that it is worth saving in agent_data? agent_data has one complication. It requires a trusted runner which means agents that use it won't work with a local runner (i.e. near cli) with a remote hub.

I think we can change that to be out of the agent data, I'll try another approach to store that data. This data it is needed when retrieving the server that a tool belongs to, but probably I can remove it from the agent_data

PR #1085, opened 2025-04-07
PostgreSQL PR for hub codebase 

zavodil:


ewiner (approved):
Great work!  It was a surprisingly straightforward review given the magnitude of the change.

hub/api/v1/sql.py
```diff
@@ -77,31 +79,36 @@ class VectorStoreFile(BaseModel):
...
+
+        self.db = psycopg.connect(
             host=getenv("DATABASE_HOST"),
             user=getenv("DATABASE_USER"),
             password=getenv("DATABASE_PASSWORD"),
-            database=getenv("DATABASE_NAME"),
-            autocommit=True,
+            dbname=getenv("DATABASE_NAME"),
+            autocommit=self.db_autocommit,
+            options="-c search_path=" + getenv("DATABASE_SCHEMA"),
```
gabehamilton, line 91: If this is `public` can we add a default?

gabehamilton, line 91: If this is public can we add a default?
```suggestion
            options="-c search_path=" + getenv("DATABASE_SCHEMA", "public"),
```

gabehamilton:
Do we need to all install `pgvector` to enable the `VECTOR` type? Maybe a follow on PR with updated README.md?

gabehamilton (approved):

arktoswb:
No longer in our plans.

Issue #1113, opened 2025-04-14
[FEAT] Add mutagen and encryption to standard framework for agent processing


jcarbonnell:
### Background

**Feature Request: Add `mutagen` and `cryptography` to standard framework**

**Description**:
I’m building a [Decentralized File Storage manager]( https://github.com/jcarbonnell/DFS_manager) as a team of NEAR AI agents, requiring metadata extraction and file encryption so far. The `mutagen` library is needed for extracting metadata from files, and `cryptography` (specifically Fernet) is needed for secure file encryption before IPFS upload. These features are critical for file processing in my agentic solution aiming to advocate the use of Decentralized System Networks as a storage for NEAR dApps.

**Use Case**:
- `mutagen`: Extracts metadata to log in threads and pass to other agents for blockchain transactions.
- `cryptography`: Encrypts files with a key, aligning with secure file sharing design patterns.

**Versions**:
- `mutagen`: Latest stable
- `cryptography`: Latest stable

**Reference**:
- Mutagen: https://github.com/quodlibet/mutagen
- Cryptography: https://github.com/pyca/cryptography

Thank you for considering this request!

### User Story

**Previous Attempts**:
I tried importing these libraries in my agents (`devbot.near/upload-agent`, `devbot.near/storage-agent`), but these direct uploads caused the agents to fail, while those agents have been working since I removed the libraries import.

**Impact**:
Adding these will enable richer file processing for NEAR AI agents, benefiting use cases like decentralized storage and similar applications.

### Acceptance Criteria

**Framework**:
Please add to the `standard` framework (`requirements-standard.txt`), as `minimal` is too limited, and `agentkit` is unrelated to my needs.

PR #1115, opened 2025-04-15
adding cryptography, mutagen, pillow, torch to standard libraries

jcarbonnell:
https://github.com/nearai/nearai/issues/1113

### DIFF
--- a/aws_runner/frameworks/requirements-standard.txt
+++ b/aws_runner/frameworks/requirements-standard.txt
@@ -117,10 +117,6 @@ setuptools >= 21.0.0
 # https://github.com/jd/tenacity
 tenacity == 9.0.0
 
-# PyTorch for machine learning
-# https://github.com/pytorch/pytorch
-torch==2.6.0
-
 # Twitter API wrapper
 # https://github.com/tweepy/tweepy
 tweepy >= 4.14.0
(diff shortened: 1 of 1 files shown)


aws_runner/frameworks/requirements-standard.txt
```diff
@@ -105,6 +117,10 @@ setuptools >= 21.0.0
 # https://github.com/jd/tenacity
 tenacity == 9.0.0
 
+# PyTorch for machine learning
+# https://github.com/pytorch/pytorch
+torch==2.6.0
```
gabehamilton, line 122: Torch is huge (5-6GB) which causes problems to pop up in many areas over time. We don't want that in the standard framework.

jcarbonnell: alright, shall I modify this PR, or cancel it and create a new one?
or could you do it on your side and merge the rest of the PR?

other question: if not it the standard framework, is there a place where to add torch to use it in an agent?
or how do you usually call HF-hosted models into agents without torch?

gabehamilton: Sure, I'll take out that lib.

gabehamilton: I believe our current process for custom models is to upload them to Fireworks as public models, https://docs.fireworks.ai/models/uploading-custom-models, then using them from there. @arktoswb is that correct?

gabehamilton (approved):

Issue #698, opened 2025-01-11
User attachments


gabehamilton:
Allow users to attach files to their messages. 

Proposed details:
    Show user attachments in chat stream.
    Continue to show output files in the output section but do not show input file there.

Notes:

https://app.near.ai/agents/flatirons.near/common-tool-library/latest/source has prompts for processing audio, pdf, and image files.

Messages support `attachments` when creating a message with a post to `/threads/{thread_id}/messages`

zavodil:
Shall we limit max attachment(s) size?

gabehamilton:
I think we were running into a 1MB limit with registry files (possible due to nginx) so worth some testing. 

Metadata flag is a great idea: `allow_message_attachments` and `supported_attachment_types`? 
The latter would be for specifying just text & markdown, or everything, etc

gabehamilton:
Fireworks also supports an interesting inlining mode https://docs.fireworks.ai/firesearch/inline-multimodal

calebjacob:
This sounds like a high priority for us to knock out @gabehamilton. Sounds like we have some backend work that needs to happen for our thread message endpoints.

CC @arktoswb @gagdiez

PR #1117, opened 2025-04-16
fix: correctly convert input `content`

gagdiez:
[`create_message`](https://github.com/nearai/nearai/blob/b3cf86efae03e58ea47ce8dbf85afbb4a7d47493/hub/api/v1/thread_routes.py#L377) receive as input a `MessageCreateParam` which can [either be a `string` or a list of pre-determinated objects](https://github.com/nearai/nearai/blob/main/hub/api/v1/thread_routes.py#L223), one of which is a `TextContentBlockParam`

Currently, we are only handling the logic for when the input is a `string`, which we correctly convert into an output `Message`. However, the code breaks when the user inputs a `TextContentBlockParam`, which is very useful in situations such as sending text and an image in the same message.

This fix handles the case for when a `TextContentBlockParam` is input, correctly converting it into a valid output `Messages`

### DIFF
--- a/hub/api/v1/models.py
+++ b/hub/api/v1/models.py
@@ -308,7 +308,8 @@ class Message(SQLModel, table=True):
             if isinstance(self.content, Iterator):
                 self.content = [
                     TextContentBlock(text=Text(value=content["text"], annotations=[]), type="text")
-                    if content["type"] == "text" else content
+                    if content["type"] == "text"
+                    else content
                     for content in self.content
                 ]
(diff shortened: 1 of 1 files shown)


gagdiez:
This is a first step necessary to make our frontend be able to handle images (e.g. send a message alongside an image asking "what do you see here?"), related to https://github.com/nearai/nearai/issues/698

arktoswb (approved):

PR #1119, opened 2025-04-16
Message Attachments

calebjacob:
Closes: https://github.com/nearai/nearai/issues/698

Loom demo: https://www.loom.com/share/1c8260cf039b4944877d21ee870fea40

By default, agents will have attachments disabled. The developer can opt into allowing file attachments to be sent to their agent by configuring metadata:

```json
{
  "details": {
    "agent": {
      "allow_message_attachments": true,
      "allow_message_attachments_accept_mime_types": [
        "image/*",
        "application/pdf"
      ]
    }
  }
}
```

If `allow_message_attachments_accept_mime_types` is not defined, all file types are allowed by the UI. It's worth noting that the hub API may throw an error due to it also only accepting specific file types. For example, the hub API currently throws an error when trying to upload a `.svg` file.


hub/demo/src/lib/models.ts
```diff
@@ -146,7 +163,21 @@ export const entryModel = z
 export const entriesModel = z.array(entryModel);
 
 export const fileModel = z.object({
+  bytes: z.number(),
```
gagdiez, line 166: The `Source` tab now breaks because files such as `agent.py` and `metadata.json` might not have many of these fields

e.g. http://localhost:3000/agents/gagdiez.near/hello-vision/latest/source?file=agent.py

Adding a bunch of `optional` fixed it for me:

```
  bytes: z.number().optional(),
  created_at: z.number().optional(),
  expires_at: z.number().nullable().optional(),
  filename: z.string(),
  id: z.string().optional(),
  object: z.string().optional(),
  purpose: z.enum([
    'assistants',
    'attachments',
    'batch',
    'fine-tune',
    'vision',
  ]).optional(),
  status: z.enum(['uploaded', 'processed', 'error']).optional(),
  status_details: z.string().optional(),
```

calebjacob: @gagdiez Thanks for catching that! I just pushed a fix.

gagdiez:
This works on my side, now we only need #1121 to be merged so agents can see the attachments!

gagdiez (approved):

PR #1125, opened 2025-04-22
Feat/enhance integration docs

### LINKED ISSUE #993: Wrong nonce format for signing messages in "Integrate NEAR AI assistant" docs
In the documentation for integrating NEAR AI into your own apps:

https://github.com/nearai/nearai/blob/main/docs/assistants/integrate.md

The code snippet below for getting an authorization token does not work because the nonce is in the wrong format.

```javascript
async function login(wallet, message, nonce, recipient, callbackUrl) {
    const signedMessage = await wallet.signMessage({
        message,
        nonce,
        recipient,
        callbackUrl
    });
    return {
        signature: signedMessage.signature,
        accountId: signedMessage.accountId,
        publicKey: signedMessage.publicKey,
        message,
        nonce,
        recipient,
        callbackUrl
    };
}

// Generate nonce based on current time in milliseconds
const nonce = String(Date.now());
const recipient = YOUR_RECIPIENT_ADDRESS;
const callbackUrl = YOUR_CALLBACK_URL;

// Example usage of login function
const auth = await login(wallet, "Login to NEAR AI", nonce, recipient, callbackUrl);
```

The `wallet.signMessage()` function expects a 32-byte buffer, but in the docs the nonce is a string.

I found this issue while working on a typical `create-near-app` project, and getting this error when calling the function:
```
Error: Invalid nonce. It must be a Buffer with a length of 32 bytes.
    at WalletModules.validateSignMessageParams (index.js:2637:13)
    at WalletModules.eval (index.js:2668:12)
    at Generator.next (<anonymous>)
    at eval (index.js:50:71)
    at new Promise (<anonymous>)
    at __awaiter (index.js:46:12)
    at wallet.signMessage (index.js:2664:36)
    at Wallet.signMessage (near.js:196:27)
```

### Solution

Looking at the demo frontend here:

https://github.com/nearai/nearai/blob/f9562fb697eed6f8511eff36be413c8926694147/hub/demo/src/lib/auth.ts#L164

The nonce is correctly formatted here:

```javascript
/**
 * Generates a nonce, which is current time in milliseconds
 * and pads it with zeros to ensure it is exactly 32 bytes in length.
 */
export function generateNonce() {
  const nonce = Date.now().toString();
  return nonce.padStart(32, '0');
}
```

So the documentation should mention that you need to format the nonce to 32 bytes, and also it needs to be passed as a buffer. I used the following in my own project to get this working:

```javascript
const nonce = Date.now().toString().padStart(32, '0');
const nonceBuffer = Buffer.from(nonce);
```

Then I pass `nonceBuffer` into `signMessage` to get the auth token.

thisisjoshford:
Fixes #993 

While working on the above fix, I discovered that the Integrations section required an update. This PR:

- Adds agent embedding to this section as primary go-to for easy agent integration
- Consolidates integration docs into one article
- Adds redirects for article relocation


docs/agents/integration.md
```diff
@@ -0,0 +1,261 @@
...
+    
+    - Clicking the `share` icon from your agent page and select `<embed>`
+
+    ![agent embed snippet](../assets/agents/agent-embed.png)
+
+### Customizing the Embed
+
+There are three ways to customize the appearance and behavior of your embedded agent:
+
+1. Basic iframe attributes
```
bucanero, line 44:
```suggestion
1. Basic `<iframe>` attributes
```

bucanero (approved):

PR #1127, opened 2025-04-23
Add option to decide how to decode files

gagdiez:
Right now our platform assumes that users will only try to use text files, and thus tries to decode every single file using `utf-8`

Now that we added the ability to use attachments, it is likely that agents will interact with files that are not text, for example images or audio

This commit adds a `decode` field to all "read" operations, trying to keep as much compatibility as possible with the current implementation. This is, by default we try to decode the file as utf-8, but this can be skipped

### DIFF
--- a/nearai/agents/environment.py
+++ b/nearai/agents/environment.py
@@ -680,7 +680,7 @@ class Environment(object):
 
         self.list_files_from_thread = list_files_from_thread
 
-        def read_file_by_id(file_id: str, decode: str | None = "utf-8"):
+        def read_file_by_id(file_id: str, decode: Union[str, None] = "utf-8"):
             """Read a file from the thread."""
             content = hub_client.files.content(file_id).content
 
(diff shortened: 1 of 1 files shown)


nearai/agents/environment.py
```diff
@@ -958,22 +961,19 @@ def get_agent_temp_path(self) -> Path:
...
-        print(f"Read file {filename} from local path: {local_path}")
+        print(f"Reading file {filename} from local path: {local_path}")
         if os.path.exists(local_path):
             try:
                 with open(local_path, "rb") as local_path_file:
                     local_file_content = local_path_file.read()
-                    try:
-                        # Try to decode as text
-                        file_content = local_file_content.decode("utf-8")
-                    except UnicodeDecodeError:
```
gagdiez, line 974: This might be the only "problematic" change, since before we were trying to decode into `utf-8`, and otherwise reading the file as `bytes`

Now that the user can explicitly add how to decode the file, we do not need to assume what to do. Either it works with the decided "decoding scheme" or it fails.

I honestly wonder how many agents anyway will be affected by this, since anyway people did not have many ways to upload / interact with files

gabehamilton: Can we fallback to utf-8 decoding?

gagdiez: sorry, I did not understand. Right now the default value for decoding is "utf-8"

What you are suggesting is trying to decode, and in case of error try to decode one last time but with "utf-8"?

arktoswb (approved):

PR #1145, opened 2025-05-08
Bring back UnicodeDecodeError handling

arktoswb:
Revert of api_client.py changes in https://github.com/nearai/nearai/pull/1136
Need to remember to review autogenerated changes under openapi_client.


gagdiez:
@arktoswb I see that you are adding an `except: return data as is`, where is this needed? does anything break without it?

arktoswb:
> @arktoswb I see that you are adding an `except: return data as is`, where is this needed? does anything break without it?

It's a revert of api_client.py changes in #1136
It broke your attachments agent, so I thought, it was connected to your changes to support attachments. But today I remembered that we had this code for a while. I need to remember to review the autogenerated code under openapi_client.

thisisjoshford (approved):

Issue #1150, opened 2025-05-12
[FEAT] Debug Button


gagdiez:
### Background

Currently, it is very hard for users to debug their live agents, as they need to both set the `logs` button and setup the `DEBUG` variable in the frontend

This could be easily replaced by a simple `debug` button, that does this automatically for the user

### User Story

As a user, I would like to easily debug my agent

### Acceptance Criteria

- [ ] Implement a button that triggers the `debug` behaviour

PR #1153, opened 2025-05-14
feat: add switch to toggle between false and true

matiasbenary:
https://github.com/nearai/nearai/issues/1150
Never show the variable `Debug`.
When the switch is toggled off.Don't delete variable , only change its value to `false`

When the switch is on
Captura desde 2025-05-13 21-38-27
When the switch is off:
Captura desde 2025-05-13 21-48-26

### DIFF
--- a/hub/demo/src/components/EntryEnvironmentVariables.tsx
+++ b/hub/demo/src/components/EntryEnvironmentVariables.tsx
@@ -10,6 +10,7 @@ import {
   Input,
   InputTextarea,
   SvgIcon,
+  Switch,
   Text,
   Tooltip,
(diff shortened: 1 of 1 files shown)


hub/demo/src/components/EntryEnvironmentVariables.tsx
```diff
@@ -52,10 +53,23 @@ export const EntryEnvironmentVariables = ({
     entry,
     excludeQueryParamKeys,
   );
+  const [debug, setDebug] = useState(false);
```
arktoswb, line 56: Can you please change default value of DEBUG to true

PR #1163, opened 2025-05-19
feat: hide debug files when in debug mode

matiasbenary:


hub/demo/src/components/AgentRunner.tsx
```diff
@@ -686,6 +692,9 @@ export const AgentRunner = ({
     if (showLoadingPlaceholder) return <PlaceholderSection />;
     return null;
   }
+  const outputFiles = showLogs
+    ? files.outputs
```
denbite, line 696: would you mind updating to use `useMemo` here so we don't iterate over array of files on each render

```jsx
const outputFiles = useMemo(() => {
// the logic here
}, [showLogs, files.outputs.length]);
```

gagdiez:
@arktoswb instead of having two buttons: `debug` and `show logs`, we use the already button `show logs` to either show or hide all logs (including log files)

most users will not want to have `logs` nor `log files` on by default, they will simply want to interact with the agents

arktoswb:
> most users will not want to have logs nor log files on by default, they will simply want to interact with the agents

Most users would prefer to have logs present when their agent run fails.
I think, `debug` button will be more useful than `show logs` button.

PR #1183, opened 2025-06-06
feat: replace show logs button with debug button

matiasbenary:


matiasbenary:
I didn’t make any changes to the Python code
image
Captura desde 2025-06-06 21-03-34

arktoswb:
Deployments are failing. Merge the latest `main` to fix

matiasbenary:
> Screenshot 2025-06-09 at 10 03 10 AM
> The chat is overfilled with debug messages in debug mode. If you are removing a logs button, make it always off. We'll want logs only on the right, and not in the chat window.

Should I merge this PR with the one in the link?
https://github.com/nearai/nearai/pull/1163

arktoswb:
> > Screenshot 2025-06-09 at 10 03 10 AM
> > The chat is overfilled with debug messages in debug mode. If you are removing a logs button, make it always off. We'll want logs only on the right, and not in the chat window.
> 
> Should I merge this PR with the one in the link? #1163

No, that PR is removing log files. Let me give you some context. Addition of log files have brought us positive changes:
1. No more "Is agent hub down?" messages. The agent hub is obviously not down if you see progress happening in log files.
2. Much less requests on why the agent is not working. Again, the cause is already there, in log files.
3. It gave me a lot of logs data for building analytics.
4. It gave me detailed latency data on each api call.

The downside was that execution speed was significantly increased: 20 seconds or even more.
On Friday I deployed speed improvements that bring down the run time to ~4 seconds on debug_mode and ~3 seconds on release mode for a trivial agent.

Current state of the agent chat window:
1. Debug mode is on by default, but it's not obvious.
2. There are log files visible on the right side of the chat window (very important feature).
3. There is a button to show log messages inside chat window. I personally don't find it useful at this moment.

There was a negative feedback from developers about debug mode being on by default. This is because debug mode is increasing run time. I already addressed it by significantly improving execution time of an agent on debug mode: from ~20 seconds to ~4 seconds for a trivial agent.

You are working on the problem of debug_mode being 'on' by default and being it not obvious for agent developers. Going forward:
1. Log files on the right from the chat window should be visible by default, because it's a very important feature, and it successfully addressed some of our problems. I would prefer for a debug_mode to stay 'on' by default when interacting with agents through the hub: either assumed 'on' if not given in `environment.py` (as it is happening right now), or explicitly passed `DEBUG_MODE=True` by a frontend (probably, a better approach)
2. Debug mode button on the right from the chat window will also be extremely useful to make it obvious that debug_mode is on by default. Disabling it should pass `DEBUG_MODE=False` to an agent run, which is what you are doing in this PR, I think.
3. I don't care what happens to the logs button and log messages inside chat window. I personally don't find this feature either easy to discover or useful. We should definitely not have it 'on' by default.

arktoswb (changes requested):
No changes since my last comment. Please read my comments.

gagdiez:
After my last commit, the frontend now behave as follows:

- there is a single DEBUG button (no more `show log` buttons)
- DEBUG is `true` by default (i.e. if no environmental variable named `DEBUG` is set)
- when DEBUG is `false`, the files `runner_log`, `system_log` and `chat_history_log` are hidden

No matter if DEBUG is true or false, logs are now always hidden on the chat (i.e. the `show logs` button functionality was removed)

PR #1198, opened 2025-09-12
feat: support for firework's workflow models in /images/generations

laciferin2024:
- [x] Add first-class support for Fireworks workflow-based image models (Flux/Kontext) to /v1/images/generations.
- [x] Implements request submission + polling to return a base64-encoded image in an OpenAI-style response.
- [x] Preserves existing Playground image generation paths.
- [x] Support for 4 image generation models added (
            "accounts/fireworks/models/flux-1-dev-fp8",
            "accounts/fireworks/models/flux-kontext-max",
            "accounts/fireworks/models/flux-1-schnell-fp8",
            "accounts/fireworks/models/flux-kontext-pro",
        )
        
- [x] /images/generations endpoint relevant cases documented on `images.generations.doc.http` 

- [x] Added support for DB Port configuration for devs like me who enjoys working on multiple projects simultaneously.


laciferin2024:
I would recommend using httpYac to execute *.http files

